PR #5, opened 2022-08-25
Expose generated prisma types in db package

ReinhardtR:
This is useful for importing stuff like enums from the schema.

### DIFF
--- a/packages/db/prisma.ts
+++ b/packages/db/prisma.ts
@@ -1,0 +1,1 @@
+export * from "@prisma/client";


juliusmarminge:
Not too keen on this as default, I almost never need to import my Prisma types so it would just increase bundle size of the app I think?

ReinhardtR:
That's fair, but will it be included in the bundle if you don't import from it? 🤔

juliusmarminge:
Maybe it's treeshaked, would have to analyze the bundle to say for sure

PR #20, opened 2022-09-19
trpc versions from alpha to beta 📦

azezsan:
I updated trpc versions to the beta versions and modified trpc utils files in both Nextjs and Expo to be compatible according to trpc's 10.x docs


azezsan:
I tested on: 
PC for nextjs and expo with android emulator 
Linux wsl for nextjs only
both worked with no problems

apps/nextjs/src/utils/trpc.ts
```diff
@@ -10,12 +11,16 @@ const getBaseUrl = () => {
...
+      links: [
+        httpBatchLink({
+          url: `${getBaseUrl()}/api/trpc`,
+        }),
+      ],
     };
   },
-  ssr: false,
-});
+  ssr: true,
```
juliusmarminge, line 25:
```suggestion
  ssr: false,
```

juliusmarminge (commented):
Thanks, i think you need to make some changes to the API package too though

juliusmarminge (approved):

azezsan:
> Thanks, i think you need to make some changes to the API package too though.
> 
> EDIT: nvm the ones i thought of were already done

happy to help ^^

PR #21, opened 2022-09-20
Include note about pnpm

ImBaedin:
Figured this might be helpful for others after I spent an hour fighting with modules not resolving.

Tried to mirror the formatting of the docs best I could.


juliusmarminge:
I think we can actually bump this above the deployment guide?

ImBaedin:
> I think we can actually bump this above the deployment guide?

I'll update it this evening.

PR #27, opened 2022-10-03
switch to pnpm

juliusmarminge:


turbo.json
```diff
@@ -10,6 +10,8 @@
     "dev": {
       "cache": false
     },
+    "clean": {},
+    "//#clean": {},
```
charkour, line 14: Thanks for this awesome starter project! I'm wondering, what is this syntax: `//#clean`? I couldn't find it on the turbo docs. Perhaps it is pnpm specific? Thanks.

juliusmarminge: It is so that when you run turbo clean the root's clean script is also run

See the difference in `clean` and `clean:workspaces` scripts in the root package json.

https://turbo.build/blog/turbo-1-3-0#run-and-cache-scripts-from-the-root-of-your-monorepo

charkour: Thanks you, @juliusmarminge!

Issue #31, opened 2022-10-11
Unable to resolve module ./index


nrowley:
Hey guys, love what you've done with this monorepo setup and the t3 stack :)

I've been running into this issue on my project:
![Screenshot 2022-10-11 at 14 45 03](https://user-images.githubusercontent.com/47508642/195108308-0e8e334f-5645-4f03-bfc1-84f61172a743.png)

this happens when i try to run the bundled ios app using `npx expo run:ios` as well `expo run:ios`

this is happening on a clean clone of the repo as well as my personal project repo :(

would love if you could point me in the right direction.

nrowley:
I need to run it using expo run:ios as I am using a datepicker library that has native code which will not work on expo go

nrowley:
i think its similar to this issue https://github.com/expo/expo-cli/issues/4509

juliusmarminge:
Maybe @byCedric has some idea?

nrowley:
please let me know if you need more info

nrowley:
just to confirm i cloned the branch again and ran the steps in the repo README

1. pnpm i
2. changed the db provider in schema.prisma to `sqlite`
3.echo DATABASE_URL=file:./db.sqlite >> packages/db/.env
4. pnpm db-push
5. cd /apps/expo
6. npx expo run:ios


i get the following error
```bash
 Config synced
✔ Installed pods and initialized Xcode workspace.
› Planning build
› Compiling react-native Pods/DoubleConversion » bignum.cc
› Compiling react-native Pods/DoubleConversion » strtod.cc
› Compiling react-native Pods/DoubleConversion » fast-dtoa.cc
› Compiling react-native Pods/DoubleConversion » fixed-dtoa.cc
› Compiling react-native Pods/DoubleConversion » double-conversion.cc
› Compiling react-native Pods/DoubleConversion » diy-fp.cc
› Compiling react-native Pods/DoubleConversion » cached-powers.cc
› Compiling react-native Pods/DoubleConversion » bignum-dtoa.cc
› Compiling react-native Pods/DoubleConversion » DoubleConversion-dummy.m
› Packaging react-native Pods/DoubleConversion » libDoubleConversion.a
› Compiling react-native Pods/glog » demangle.cc
› Compiling react-native Pods/fmt » fmt-dummy.m
› Compiling react-native Pods/RCT-Folly » Assume.cpp
› Compiling react-native Pods/Yoga » event.cpp
[...]

PR #32, opened 2022-10-12
fix: rename expo entry file

nrowley:
this is a fix for the issue i was having in #31 

it seems that the entry point for expo needs to be index.*

so i've renamed and moved the file up out of the src folder into the root.

`npx expo run:ios` now works, notice the index.tsx loading bar, this was not there previously as index was not found...

now we are getting the expected localhost error

Screenshot 2022-10-12 at 19 07 25
Screenshot 2022-10-12 at 19 18 38

### DIFF
--- a/apps/expo/index.ts
+++ b/apps/expo/index.ts
@@ -1,0 +1,8 @@
+import { registerRootComponent } from "expo";
+
+import App from "./src/_app";
+
+// registerRootComponent calls AppRegistry.registerComponent('main', () => App);
+// It also ensures that whether you load the app in Expo Go or in a native build,
+// the environment is set up appropriately
+registerRootComponent(App);
--- a/apps/expo/index.ts
+++ b/apps/expo/index.ts
@@ -1,6 +1,6 @@
 import { registerRootComponent } from "expo";
 
-import App from "./src/_app";
+import { App } from "./src/_app";
 
 // registerRootComponent calls AppRegistry.registerComponent('main', () => App);
 // It also ensures that whether you load the app in Expo Go or in a native build,
--- a/apps/expo/src/_app.tsx
+++ b/apps/expo/src/_app.tsx
@@ -5,7 +5,7 @@ import { TRPCProvider } from "./utils/trpc";
 
 import { HomeScreen } from "./screens/home";
 
-const App = () => {
+export const App = () => {
   return (
     <TRPCProvider>
       <SafeAreaProvider>
@@ -15,4 +15,3 @@ const App = () => {
     </TRPCProvider>
   );
 };
-
--- a/apps/expo/src/_app.tsx
+++ b/apps/expo/src/_app.tsx
@@ -1,4 +1,3 @@
-import { registerRootComponent } from "expo";
 import { StatusBar } from "expo-status-bar";
 import React from "react";
 import { SafeAreaProvider } from "react-native-safe-area-context";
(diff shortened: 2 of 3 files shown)


nrowley:
@juliusmarminge

juliusmarminge:
This makes no sense though, the default entrypoint in a fresh expo app is App.js

nrowley:
yeah I was reading through these docs here and it seemed to work.

https://docs.expo.dev/guides/monorepos/

juliusmarminge:
Does any filename work if its in the root maybe?

nrowley:
no it doesn't i tried using _app.tsx it looks like it specifically looks for index.*

juliusmarminge:
Hmm thats weird, must be configurable? If it isn't i guess this Will do but i would like the file to atleast exist under `src`

nrowley:
i will see what i can do, probably need to set it in the app.json?

nrowley:
I dont think its configurable, i've tried setting the `"entryPoint":"./src/_app.tsx"` in the app.json but it didn't work

juliusmarminge:
wtf is this crap, it removes the entryPoint and sets it to index even if you have it set to something else
CleanShot 2022-10-12 at 21 11 26

nrowley:
yeah man its weird.
Im trying to App from src/_app.tsx and import it in the root at index.ts as the expo monorepo docs say...

juliusmarminge:
> yeah man its weird.
> Im trying to App from src/_app.tsx and import it in the root at index.ts as the expo monorepo docs say...

yea that's the next best thing, just registerring the app from root and doing the actual providers etc in src/_app

nrowley:
okay, that worked. i pushed the changes

apps/expo/index.ts
```diff
@@ -0,0 +1,8 @@
+import { registerRootComponent } from "expo";
+
+import App from "./src/_app";
```
juliusmarminge, line 3:
```suggestion
import { App } from "./src/_app";
```

apps/expo/src/_app.tsx
```diff
@@ -17,4 +16,4 @@ const App = () => {
   );
 };
 
-registerRootComponent(App);
+export default App;
```
juliusmarminge, line 19: `export const App` on row 8

```suggestion
```

juliusmarminge:
I can't add suggestion for row 8 since you haven't changed it but do a regular export there
`export const App = ...`

nrowley:
done

juliusmarminge (approved):
Awesome, weird issue. Might pull an issue on Expo since this seems weird if this is how it's supposed to work...

Why let me modify it if you're just gonna override it lol

nrowley:
yeah man so weird it seems like it just doesn't work at all...

juliusmarminge:
> Awesome, weird issue. Might pull an issue on Expo since this seems weird if this is how it's supposed to work...
> 
> 
> 
> Why let me modify it if you're just gonna override it lol

Will be supported in Expo 47: https://github.com/expo/expo/issues/19527

nrowley:
> > Awesome, weird issue. Might pull an issue on Expo since this seems weird if this is how it's supposed to work...
> > Why let me modify it if you're just gonna override it lol
> 
> Will be supported in Expo 47: [expo/expo#19527](https://github.com/expo/expo/issues/19527)

cool man, thanks for the help! 🚀

PR #35, opened 2022-10-19
fix: make build/dev depend on db-generate

juliusmarminge:
Dont `prisma generate` on postinstall, but instead make build/dev depend on it.

Ref: https://twitter.com/mattpocockuk/status/1582660272330018816

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -1,17 +1,19 @@
 {
   "pipeline": {
     "build": {
-      "dependsOn": ["^build"],
+      "dependsOn": ["^build", "db-generate"],
       "outputs": [".next/**", ".expo/**"]
     },
     "lint": {
       "outputs": []
     },
     "dev": {
+      "dependsOn": ["db-generate"],
       "cache": false
     },
+    "db-generate": {},
+    "db-push": {},
     "clean": {},
-    "//#clean": {},
-    "postinstall": {}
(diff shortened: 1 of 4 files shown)


turbo.json
```diff
@@ -1,17 +1,19 @@
 {
   "pipeline": {
     "build": {
-      "dependsOn": ["^build"],
+      "dependsOn": ["^build", "db-generate"],
```
mattpocock, line 4:
```suggestion
      "dependsOn": ["^build", "^db-generate"],
```

turbo.json
```diff
@@ -1,17 +1,19 @@
...
       "outputs": [".next/**", ".expo/**"]
     },
     "lint": {
       "outputs": []
     },
     "dev": {
+      "dependsOn": ["db-generate"],
```
mattpocock, line 11:
```suggestion
      "dependsOn": ["^db-generate"],
```

turbo.json
```diff
@@ -1,17 +1,19 @@
...
       "cache": false
     },
+    "db-generate": {},
```
mattpocock, line 14: You may also want to add `cache: false` here until you can figure out which files to cache.

Same with `db-push`, `clean` and `//#clean`

juliusmarminge: Alright thanks :)

PR #36, opened 2022-10-19
feat: add next-auth package

juliusmarminge:
con: doesn't work with expo yet :/


matteobad:
Hi @juliusmarminge,

thank you for working on this! I've tried this next-auth integration in one of my current projects, and I have a few problems. Before writing I checked this branch and I got the same errors:

- First a runtime error in the `apps/nextjs`
Schermata 2022-10-23 alle 14 50 55

- Second in the api package when using a `protectedProcedure` the types of the `Session` does not contain the `user.id`. It seems that the override of the `next-auth` types made in the `packages/auth` is not in place in other packages

I'm not an expert of neither typescript or monorepos, but happy to help in any way possible :)
So let me know if I can do anything to help make this branch work

matteobad:
Update:

- I was able to solve the first problem by adding `@acme/auth` to the list of modules that need transpilation `withTM(["@acme/api", "@acme/auth", "@acme/db"])`
- As for the second problem, I kinda hack it by adding the `next-auth.d.ts` in the list of inclusion in the `tsconfig.json` of the importing package. Example `@acme/api tsconfig.json` -> `include: [..., "node_modules/@acme/auth/next-auth.d.ts"]`. By doing so the type for `Session` inside the `ctx` in a `protectedProcedure` contains `id` property 

The first solution can be used in the same way in this branch.
As for the second solution I'm not sure it's the right way to do it.

Le me know what do you think. Bye 👋

juliusmarminge:
> Update:
> 
> * I was able to solve the first problem by adding `@acme/auth` to the list of modules that need transpilation `withTM(["@acme/api", "@acme/auth", "@acme/db"])`
> * As for the second problem, I kinda hack it by adding the `next-auth.d.ts` in the list of inclusion in the `tsconfig.json` of the importing package. Example `@acme/api tsconfig.json` -> `include: [..., "node_modules/@acme/auth/next-auth.d.ts"]`. By doing so the type for `Session` inside the `ctx` in a `protectedProcedure` contains `id` property
> 
> The first solution can be used in the same way in this branch. As for the second solution I'm not sure it's the right way to do it.
> 
> Le me know what do you think. Bye 👋

You can file a PR into this if you want :)

matteobad:
Hi @juliusmarminge, I've created a PR on the `auth` branch with the above fixes. #48 
Let me know if there is anything that needs to be changed :)

juliusmarminge:
I believe it's in a working state right now. Feel free to scaffold from this branch and let me know if any issues arise. I'll go over the configuration a bit more before I'll merge it into main

PR #37, opened 2022-10-19
Generate prisma client outside of node modules, cache generated client

TheoBr:
I think this works

### DIFF
--- a/packages/db/.gitignore
+++ b/packages/db/.gitignore
@@ -1,0 +1,1 @@
+.generated/*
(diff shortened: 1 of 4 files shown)


juliusmarminge (approved):

TheoBr:
Turns out this is entirely broken - something like the chaos @roxkstar74 did here will be necessary probably :(

https://github.com/prisma/prisma/issues/10433#issuecomment-1264687986

PR #41, opened 2022-10-20
Prevent creation of multiple prisma clients

ImBaedin:
After running into issues with the prisma client creating multiple connections, I did some digging and found that it's a [common problem with next](https://www.prisma.io/docs/guides/database/troubleshooting-orm/help-articles/nextjs-prisma-client-dev-practices). 

This PR uses a section of code from [this issue](https://github.com/prisma/prisma/issues/5007#issuecomment-740803749) to resolve the problem by preventing the client from being recreated.

Quick note: I didn't clone this repo when starting my project, I followed the article on migrating an existing t3 app. If there is already a fix for this somewhere that I missed, my bad.

### DIFF
--- a/packages/db/index.ts
+++ b/packages/db/index.ts
@@ -1,23 +1,19 @@
 import { PrismaClient } from "./.generated/client";
 
-let prisma: PrismaClient;
-if (process.env.NODE_ENV === "production") {
-  prisma = new PrismaClient({
-    log: ["query"],
-  });
-  console.log("Production: Created DB connection.");
-} else {
-  // @ts-ignore
-  if (!global.prisma) {
-    // @ts-ignore
-    global.prisma = new PrismaClient({
-      log: ["query"],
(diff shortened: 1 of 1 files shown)


juliusmarminge:
Use the snippet we have in create-t3-app for consistency.

However I'm not sure if this is needed in monorepos? I think that's why it's not there but could have forgotten something and be completely wrong there

ImBaedin:
> Use the snippet we have in create-t3-app for consistency.
> 
> However I'm not sure if this is needed in monorepos? I think that's why it's not there but could have forgotten something and be completely wrong there

I was personally running into the issue, I could share access to the repo I was struggling with if that helps.

Like I said, it _could_ be a config issue from migrating to turbo.

I'm headed to the gym now, and can update to match the convention when I get home this evening.

juliusmarminge (approved):

Issue #42, opened 2022-10-20
Prisma broken for next when generating prisma client in non-node_modules dirs


TheoBr:
My bad :(

Broke by my PR here:
https://github.com/t3-oss/create-t3-turbo/pull/37

Error here, unable to find schema:
https://twitter.com/t3dotgg/status/1582936894991609857
![image](https://user-images.githubusercontent.com/6751787/196853258-010df0d7-3965-4c84-a016-34a7b9adffbc.png)

Tried solution here:
https://github.com/prisma/prisma/issues/10433#issuecomment-1264687986

Gets further, but new error:
```
@acme/nextjs:dev: prisma:query SELECT `main`.`Post`.`id`, `main`.`Post`.`title`, `main`.`Post`.`content` FROM `main`.`Post` WHERE 1=1 LIMIT ? OFFSET ?
@acme/nextjs:dev: prisma:query SELECT 1
@acme/nextjs:dev: prisma:query SELECT name FROM sqlite_schema
@acme/nextjs:dev:       WHERE type='table'
@acme/nextjs:dev:       ORDER BY name;
@acme/nextjs:dev:       
@acme/nextjs:dev: failed to fetch, PrismaClientKnownRequestError: 
@acme/nextjs:dev: Invalid `prisma.post.findMany()` invocation:
@acme/nextjs:dev: 
@acme/nextjs:dev: 
@acme/nextjs:dev: The table `main.Post` does not exist in the current database.
@acme/nextjs:dev:     at RequestHandler.handleRequestError (webpack-internal:///(api)/../../packages/db/.generated/client/runtime/index.js:31874:19)
@acme/nextjs:dev:     at RequestHandler.request (webpack-internal:///(api)/../../packages/db/.generated/client/runtime/index.js:31853:18)
@acme/nextjs:dev:     at async PrismaClient._request (webpack-internal:///(api)/../../packages/db/.generated/client/runtime/index.js:32853:24)
@acme/nextjs:dev:     at async eval (webpack-internal:///(api)/../../packages/api/src/router/post.ts:19:26)
@acme/nextjs:dev:     at async resolveMiddleware (file:///Users/theobrowne/Code/contrib/npm-turbo-t3/node_modules/@trpc/server/dist/router-28a2c129.mjs:190:30)
[...]

PR #44, opened 2022-10-20
Revert "Generate prisma client outside of node modules, cache generated client"

TheoBr:
Reverts t3-oss/create-t3-turbo#37

Because https://github.com/t3-oss/create-t3-turbo/issues/42


juliusmarminge (approved):
Lets hope we find a proper fix later

PR #45, opened 2022-10-20
Restore "Generate prisma client outside of node modules, cache generated client"

t3dotgg:
un-reverts t3-oss/create-t3-turbo#37

First we should fix https://github.com/t3-oss/create-t3-turbo/issues/42

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -1,5 +1,8 @@
 {
   "pipeline": {
+    "db-generate": {
+      "outputs": [".generated/**"]
+    },
     "build": {
       "dependsOn": ["^build", "^db-generate"],
       "outputs": [".next/**", ".expo/**"]
@@ -9,9 +12,6 @@
     },
     "dev": {
       "dependsOn": ["^db-generate"],
-      "cache": false
-    },
-    "db-generate": {
       "cache": false
     },
     "db-push": {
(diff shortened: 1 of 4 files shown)


juliusmarminge:
As long as my execution hours on railway doesn't run out the test to verify is that the posts are loading on the preview deployment

turbo.json
```diff
@@ -1,5 +1,8 @@
 {
   "pipeline": {
+    "db-generate": {
+      "outputs": [".generated/**"]
+    },
```
timbuckley, line 5: I'm curious, why bother hoisting `db-generate` to the top here?

TheoBr: I like having this read "in the order things happen", the other order felt arbitrary

fysp11 (approved):
Good step in the right direction :)

juliusmarminge:
Don't think this is possible until Prisma fixes their beloved relationship with the node_modules.

@mattpocock attempted to fix this for the [turborepo docs](https://turbo.build/repo/docs/handbook/prisma#7-caching-the-results-of-prisma-generate) without success.

mattpocock:
Yeah, this doesn't work when deployed to Vercel, at least.

PR #46, opened 2022-10-20
Enable Prisma Client caching by Turbo and hoist Client artifacts

ernestoresende:
This PR addresses the error `warn(prisma-client) There are already 10 instances of Prisma Client actively running` when adding Prisma to Next.js, as recommeded by Prisma on https://www.prisma.io/docs/guides/database/troubleshooting-orm/help-articles/nextjs-prisma-client-dev-practices.

Additionally, I've been able to restore turbo caching to the `generate-db` script from `packages/db` and hoist the generated artifacts from the client to the root of the project. This seems to solve the problem found with Next, where it doesn't play well with non-node_module dirs.

To reproduce and test caching behavior for the Prisma Client and applications:
1. Clone the repo and install the dependencies.
2. Run the `build` script on the root of the project. It should both generate the client, and build the `@acme/nextjs` project.
3. Run the `build` script a second time. It should be a full Turbo cache of both the Prisma client and the Next.js application.
4. Modify something on `@acme/nextjs` to invalidate it's cache. Run the `build` script again. The Prisma Client should be pulled from cache, while the application is rebuilt.
5. Modify something on the Prisma schema file. Run the `build` script again. Both the Prisma Client and the application should get rebuilt. 

I can't say for sure how well these settings are going to play with Turbo's Remote caching, and my current environment also doesn't let me experiment with the Expo project very well.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -12,7 +12,7 @@
       "cache": false
     },
     "db-generate": {
-      "cache": false
+      "cache": true 
     },
     "db-push": {
       "cache": false
--- a/.npmrc
+++ b/.npmrc
@@ -2,3 +2,4 @@
 # The symbolic links of pnpm break the rules of Expo monorepos.
 # @link https://docs.expo.dev/guides/monorepos/#common-issues
 node-linker=hoisted
+public-hoist-pattern[]=*prisma*
--- a/packages/db/index.ts
+++ b/packages/db/index.ts
@@ -1,4 +1,4 @@
-import { PrismaClient } from '@prisma/client';
+import { PrismaClient } from "@prisma/client";
 
 declare global {
   // allow global `var` declarations
--- a/packages/db/index.ts
+++ b/packages/db/index.ts
@@ -15,6 +15,6 @@ export const prisma =
         : ["error"],
   });
 
-if (process.env.NODE_ENV !== "production") global.prisma = prisma;
-
-export type { PrismaClient };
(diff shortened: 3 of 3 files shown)


turbo.json
```diff
@@ -12,7 +12,7 @@
       "cache": false
     },
     "db-generate": {
-      "cache": false
+      "cache": true
```
juliusmarminge, line 15: How will it know when to invalidate the cache when there's no output set?

ernestoresende: That's a good question. From what I've gathered from TurboRepo's docs, the hashing process is dependent on more factors than just the `outputs` option specified on the pipeline config (see https://turborepo.org/docs/core-concepts/caching#hashing). 

If i'm to make and educated guess, this is the condition it is relying on in this case:

> Hash the contents of all version-controlled files in the workspace folder or the files matching the inputs globs, if present

So the invalidation will actually happen by checking hash changes between files of the workspace (eg. `index.ts` that exports the Prisma Client, or the actual schema file). This is corroborated by local testing: 

Running `pnpm db-generate` for the first time, no caching:

image

Running `pnpm db-generate` for the second time, without modifying any files on `packages/db`, picks up cached result:

image

Running `pnpm db-generate` for the third time, now modifying something on the `index.ts` file in `packages/db`, hash does not match, cache is invalidated:

image

juliusmarminge (approved):

PR #48, opened 2022-10-26
Fix `auth` branch errors

matteobad:
Integrates #36

- Fix `nextjs` compilation error by adding `@acme/auth` to the list of nextjs transpiled modules
- Add typescript validation to next-auth `Session` augmented type in packages other than `@acme/auth`
- Add Auth example in nextjs web app

### DIFF
--- a/apps/nextjs/tsconfig.json
+++ b/apps/nextjs/tsconfig.json
@@ -1,4 +1,11 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "**/*.cjs", "**/*.mjs"]
+  "include": [
+    "next-env.d.ts",
+    "node_modules/@acme/auth/next-auth.d.ts",
+    "**/*.ts",
+    "**/*.tsx",
+    "**/*.cjs",
+    "**/*.mjs"
+  ]
 }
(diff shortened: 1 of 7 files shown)


packages/api/tsconfig.json
```diff
@@ -1,4 +1,9 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["src", "index.ts", "transformer.ts"]
+  "include": [
+    "src",
+    "index.ts",
+    "transformer.ts",
+    "node_modules/@acme/auth/next-auth.d.ts"
```
juliusmarminge, line 7: Should this be in `include`? `compilerOptions.types` seems more appropriate. Also maybe we should make the types a public entry point so we don't have to reference it through node_modules?

apps/nextjs/tsconfig.json
```diff
@@ -1,4 +1,11 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "**/*.cjs", "**/*.mjs"]
+  "include": [
+    "next-env.d.ts",
+    "node_modules/@acme/auth/next-auth.d.ts",
```
juliusmarminge, line 5: See other comment

packages/api/tsconfig.json
matteobad: That was my initial thought, but I was not able to get it to work (surely I was missing something). I'll take a look at it as soon as I can

juliusmarminge: I'll merge for now, feel free to look into that and make another PR. I'll be busy with other stuff so prob not have time for it myself

Issue #54, opened 2022-11-06
Environment variable not found: DATABASE_URL


uze:
When I run `npx prisma studio --schema ./packages/db/prisma/schema.prisma` the prisma studio shows all data in my DB correctly, but when I run `npx turbo run dev` I get:

```Message: Error in Prisma Client request: 


Invalid `prisma.post.count()` invocation:


error: Environment variable not found: DATABASE_URL.
  -->  schema.prisma:7
   | 
 6 |   provider = "postgresql"
 7 |   url      = env("DATABASE_URL")
   | 

Validation Error Count: 1
```

Do we need to add another .env in that folder or some other configuration?

juliusmarminge:
The scripts are run with dotenv that pulls the env from the next app.

Check the scripts in db/package.json

juliusmarminge:
https://github.com/t3-oss/create-t3-turbo/blob/main/packages/db/package.json

uze:
Ah yes this fixed my issue, we can close this issue! Hmm. Should we have one global .env to read from instead of it being inside just the next app?

uze:
Aka changing `"with-env": "dotenv -e ../../apps/nextjs/.env"` to `"with-env": "dotenv -e ../../.env"`

juliusmarminge:
> Ah yes this fixed my issue, we can close this issue! Hmm. Should we have one global .env to read from instead of it being inside just the next app?

No I don't think so, you should have the envs in the apps to mirror how it is when you deploy them in production.

uze:
Hmm got it, I guess the way I was thinking about it is since both apps will share the same database, authentication, etc, then there will be lots of duplicates between the environment variables for each app, thus having one single .env that is the source of truth would be beneficial. Let me know what you think, if you like it I can make the PR otherwise we can keep it as is.

uze:
> Hmm got it, I guess the way I was thinking about it is since both apps will share the same database, authentication, etc, then there will be lots of duplicates between the environment variables for each app, thus having one single .env that is the source of truth would be beneficial. Let me know what you think, if you like it I can make the PR otherwise we can keep it as is.


Hmm got it, I guess the way I was thinking about it is since both apps will share the same database, authentication, etc, then there will be lots of duplicates between the environment variables for each app, thus having one single .env that is the source of truth would be beneficial. Let me know what you think, if you like it I can make the PR otherwise we can keep it as is.

uze:
[...]

PR #55, opened 2022-11-06
feat(env): add global .env

uze:
Added a global .env for both apps to reference from instead of having it inside of the nextjs app. 

Refer: https://github.com/t3-oss/create-t3-turbo/issues/54


juliusmarminge:
Have you tested it?

uze:
> Have you tested it?

Yep! Once you in the root of the repo you simply run `cp .env.example .env` and it'll generate your .env file then you replace the `DATABASE_URL=file:./db.sqlite` with your real URL like normal and everything works between the expo + next app. 👍🏻

uze:
Demo: @juliusmarminge 
https://user-images.githubusercontent.com/24923657/200193642-7d8ef667-b304-4041-b4d3-63c5676bc30e.mov

README.md
```diff
@@ -51,7 +51,7 @@ pnpm i
 
 # Configure environment variables.
 # There is an `.env.example` in the Next.js app you can use for reference
-cp apps/nextjs/.env.example apps/nextjs/.env
```
juliusmarminge, line 54: Update the comment above this line

uze: Updated the readme @juliusmarminge

juliusmarminge (approved):

fevernova90:
@uze I believe we also need the with-env inside apps/nextjs if we're going to use global .env

Screenshot 2022-12-27 at 9 13 33 AM

PR #66, opened 2022-11-11
Update .env.example comments

asobirov:


.env.example
```diff
@@ -1,12 +1,12 @@
-# Since .env is gitignored, you can use .env-example to build a new \`.env\` file when you clone the repo.
+# Since .env is gitignored, you can use .env.example to build a new \`.env\` file when you clone the repo.
 # Keep this file up-to-date when you add new variables to \`.env\`.
 
 # This file will be committed to version control, so make sure not to have any secrets in it.
 # If you are cloning this repo, create a copy of this file named \`.env\` and populate it with your secrets.
 
 # We use dotenv to load Prisma from Next.js' .env file
 # @see https://www.prisma.io/docs/reference/database-reference/connection-urls
-DATABASE_URL=file:./db.sqlite
```
juliusmarminge, line 9: SQLite here is fine. Should be quick to get the app running in the first place and SQLite works for that

PR #82, opened 2022-11-25
Fix SafeAreaView import to work correctly on Android

marioteik:
SafeAreaView imported from react native is not working on Android. The correct is to import it from react-native-safe-area-context.

### DIFF
--- a/apps/expo/src/screens/home.tsx
+++ b/apps/expo/src/screens/home.tsx
@@ -1,13 +1,13 @@
 import React from "react";
 
 import {
-  SafeAreaView,
   Text,
   TextInput,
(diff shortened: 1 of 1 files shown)


juliusmarminge:
Can you post some screenshots?

marioteik:
> Can you post some screenshots? I don't have an android to test on

Sure!

With the correct import: 
image

With the old import:
image

In my opinion or you guys use `SafeAreaView` from `react-native` and make sure it is working for both Android and iOS or remove SafeAreaProvider from `src/_app.tsx`.

juliusmarminge (approved):
LGTM thanks for noticing

PR #86, opened 2022-12-06
fix: point to right index.ts

dominikvrbic:
the package.json in the auth package was pointing to the wrong index.ts


juliusmarminge (approved):
🙏🏼

PR #96, opened 2022-12-27
feat(ci): skip nextjs env validation

thlmenezes:
Use environment variable to skip validations on `CI.yml` workflow, or is there a reason to validate env schema there?


juliusmarminge (approved):
makes sense 👍🏼 

a case when it would make sense is if you add e2e tests with playwright for example but since we dont have that this makes sense

PR #128, opened 2023-01-25
chore: Add SKIP_ENV_VALIDATION notice to README

albertdbio:
The SKIP_ENV_VALIDATION was introduced in [1824e40](https://github.com/t3-oss/create-t3-turbo/commit/1824e40241db58f3a85d574816075eaa0a547397).

I updated the README to provide a notice that it must be set to true when deploying to production.
Otherwise, a deployment to Vercel will fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ Let's deploy the Next.js application to [Vercel](https://vercel.com/). If you ha
 
 > The install command filters out the expo package and saves a few second (and cache size) of dependency installation. The build command makes us build the application using Turbo.
 
-2. Add your `DATABASE_URL` environment variable.
+2. Add your `DATABASE_URL` environment variable. Ensure `SKIP_ENV_VALIDATION` is set to `true` when deploying to production.
 
 3. Done! Your app should successfully deploy. Assign your domain and use that instead of `localhost` for the `url` in the Expo app so that your Expo app can communicate with your backend when you are not in development.
 


juliusmarminge:
Thats not correct.

1. The flag has been there since create-t3-app had it

2. You should generelly be ok not caring about it if you enter your envs at build time. Building in a docker container is an example of when you don't do this in which case you need the flag; https://create.t3.gg/en/deployment/docker#3-create-dockerfile

---

What issue are you running into which makes the build fail for you? Where are you deploying? The Next.js app on this repo is deployed on Vercel with fully working envs

albertdbio:
You are correct.
The issue was that I would get a missing env variable error when deploying to Vercel (see screenshot).
My mistake was not including NEXTAUTH_SECRET in the Vercel deployment page. I was only including DATABASE_URL as specified in the README.
Just trying to get a deep understanding of Turborepos, so followed line by line.

Screenshot 2023-01-25 at 2 54 49 PM

juliusmarminge:
> You are correct. The issue was that I would get a missing env variable error when deploying to Vercel (see screenshot). My mistake was not including NEXTAUTH_SECRET in the Vercel deployment page. I was only including DATABASE_URL as specified in the README. Just trying to get a deep understanding of Turborepos, so followed line by line.
> 
> Screenshot 2023-01-25 at 2 54 49 PM

Ah yea might not have updated the README since I added the auth package

PR #131, opened 2023-01-28
test: multiple prisma clients

juliusmarminge:


ecwyne:
VERY curious to see how this works out for you. I've run into issues with exactly this

sohcah:
Super excited looking at how t3 turbo is progressing, and this PR looks interesting.

I might be wrong here, but this method with the pnpm overrides seems like an over-complication.
It seems like most of this is just to deal with prisma putting things in node_modules?

In one of my own monorepos, I've found what I believe is a better alternative to integrating Prisma which avoids any node-modules modifications with a wrapping package (eg. `@acme/db1`):
- Setting Prisma's `output` option to `./.prisma/client` to  place the built Prisma client in a subdirectory of `@acme/db1`
- Setting `@acme/db1` to have `build` and `postinstall` scripts running `prisma generate` (which can then be triggered by turborepo as `@acme/db1` is a dependency)
- Having `"main": ".prisma/client"` in the `package.json` of `@acme/db1`
Then doing the same thing with eg `@acme/db2`

This allows importing `PrismaClient` from `@acme/db1` and from `@acme/db2`, whilst completely avoiding overrides and modifications to node modules.

PR #135, opened 2023-01-29
feat: add env package

### LINKED ISSUE #87: feat: better way to validate / import env vars
Hello! I am moving a create-t3-app project to a turborepo based on this template, and I notice that the env checker remains in the next folder, contrary to what I would have thought since now it wouldn't be fully checking the env variables for other apps or packages. It would be great to add a new package that checks all as a dependency in the TurboRepo workflows, similar to the env folder in @acme/next.

juliusmarminge:
Closes #87 

Adds `@acme/env` package to have typesafe env's across the monorepo.

## Usage

1. All envs are added in a single root `.env`. When deploying apps, make sure the deployment has all required envs for the apps and the packages it uses

3. Define your schema (as well as destruct process.env) like you would in a normal `T3 App` in `packages/env/index.ts`.

```ts
export const server = z.object({
  NODE_ENV: z.enum(["development", "test", "production"]),
  TEST: z.string().transform((s) => s.length),
});
```

4. Then, in every app/pkg which needs to use some env var, pull it in using `getEnv`

```ts
import { getEnv } from "@acme/env";

export const env = getEnv(["NODE_ENV"]);
env.NODE_ENV;
//  ^? development | test | production

// ------------------------------------
import { getEnv } from "@acme/env";

export const env = getEnv(["TEST", "NODE_ENV"]);
//           ^? { TEST: number, NODE_ENV: development | test | production }
```

## Cons

- Not sure what the performance hits are - we're mapping over all the keys so for large .env files this might make an impact?

- As i set it up now, the Next.js app doesn't use any `env` explicitely - all that usage is in the auth/api package - so in the current setup we aren't getting validation on build time but the errors come at runtime. This could be solved by just pulling all the AUTH/API envs into the nextjs app too ofc, but it's not the best ergonomics? You could also choose to export the `@acme/auth`'s env and import that in the nextjs app?


adambarito:
Overall this might be the best way to go. A couple thoughts:

- It's definitely cleaner DX to have a single getEnv in packages/env like this. My reasoning in having each package/app pass its own schemas (picked from the server and client schemas defined in the env package) to the env package is that way at runtime zod is only parsing the necessary variables for each package, instead of parsing all variables and then sending the ones each package requests. I honestly don't know the performance hit from zod parsing env variables every time they're called - maybe that adds up for larger repos with bigger envs, in which case it might be better to minimize what zod is doing at runtime.

- Not sure about best way to do checks at build time. Since the Next.js app is already compiling the auth and api packages at build time, would it be any different to import all the auth and api env variables directly into the app? I can check tomorrow, but I know I've had Next.js builds fail from a missing env variable in API, even though it wasn't directly imported.

LucasAlda:
This is great! It solves the worst dx on create-t3-turbo at the time. I agree that losing env validation during build time is an issue, but it could be solved by adding a validation that runs on the `pnpm run build` or `pnpm run dev` scripts in the package. json and turbo workflow. Would that resolve all of these problems, or am I wrong (I didn't use TurboRepo other than for a few experiments)?

dominikwozniak (approved):

juliusmarminge:
I'll keep this in draft for awhile. This will mess up caching in that one env variable change will most likely bust the entire turbo cache

grmkris:
hello really nice idea, here are some of my thoughts:

- env package should bo scopable by different apps
- packages shouldn't depend on environment variables

juliusmarminge:
> * packages shouldn't depend on environment variables

Yea this is the plan to refactor the packages so they are dry

vladi-strilets:
I was thinking also about having a `.ts` file per application, something like `envs.ts`
where you can define multiple zod schemas (for client + server environments)
```
// envs.ts or apps/nextjs/envs.ts

export const clientEnvs = z.object({})
export const serverEnvs = z.object({})
```

and later:

1. being able to validate the schemas before you run your app, e.g 
   ```
   "dev": "envs check && next dev"
   ```
2. infer the type of the `.ts` zod schema and continually make it synchronized. [i18next has something similar but with JSON files](https://www.i18next.com/overview/typescript#create-a-declaration-file)
   ```
   import { clientEnvs } from "envs"
   
   clientEnvs.   // <- typesafed list of options
   ```

But that sounds more like a new library

Tobjoern:
Are there any updates on this PR?

PR #136, opened 2023-01-29
feat: add expo router

PiotrekPKP:
Adds Expo Router as it's already in RC.


apps/expo/tsconfig.json
```diff
@@ -4,5 +4,5 @@
     "strict": true,
     "jsx": "react-native"
   },
-  "include": ["src", "index.ts", "app.config.ts"]
+  "include": ["app", "utils", "types", "index.ts", "app.config.ts"]
```
juliusmarminge, line 7: can we not use src with expo-router 👀 👀

PiotrekPKP: We technically can, but we cannot use src + app dir (expo router looks for app dir to find pages)
That's why I thought that having both app dir and src dir just for types and utils would be kind of weird

PiotrekPKP: Also, I have totally no clue how changing files in expo app makes NextJS's lint fail haha

juliusmarminge: > We technically can, but we cannot use src + app dir (expo router looks for app dir to find pages)
> 
> That's why I thought that having both app dir and src dir just for types and utils would be kind of weird

You should open an issue about that on their part

PiotrekPKP: > > We technically can, but we cannot use src + app dir (expo router looks for app dir to find pages)
> > That's why I thought that having both app dir and src dir just for types and utils would be kind of weird
> 
> You should open an issue about that on their part

Yeah, I definitely will

PiotrekPKP: Issue is open: https://github.com/expo/router/issues/231

PiotrekPKP:
Until the `expo-router` issue is solved, we'd have to use this folder structure. 

Also, I'm not sure why NextJS started having problem with linting and why I had to add `next` dependency to `eslint-config-next`, but it works so... :shrug:

packages/config/tailwind/index.js
```diff
@@ -1,6 +1,12 @@
 /** @type {import('tailwindcss').Config} */
 module.exports = {
-  content: ["./src/**/*.{ts,tsx}", "./src/_app.tsx"],
+  content: [
+    // NextJS
+    "./src/**/*.{ts,tsx}",
+    // Expo
+    "./app/*.{ts,tsx}",
+    "./app/**/*.{ts,tsx}",
+  ],
```
juliusmarminge, line 9: Just include `src/**` here then expo Will have to extend the preset and handle "it's special case".

juliusmarminge (commented):
I still think the rest of the files should be in src even though expo router doesnt support it (yet)

packages/config/tailwind/index.js
PiotrekPKP: Alright, will do it today

PiotrekPKP: Implemented :)

juliusmarminge:
Question though, is this the proper way to grab params from dynamic paths?

Should we not get something from props? I only get a fixed string `[id]` and not the unique id o the "request"

CleanShot 2023-02-08 at 21 01 33

apps/expo/app/index.tsx
```diff
@@ -1,21 +1,23 @@
...
   onDelete: () => void;
-}> = ({ post, onPress, onDelete }) => {
+}> = ({ post, onDelete }) => {
+  const router = useRouter();
+
   return (
     <View className="flex flex-row rounded-lg bg-white/10 p-4">
       <View className="flex-grow">
-        <TouchableOpacity onPress={onPress}>
+        <TouchableOpacity onPress={() => router.push(`/post/${post.id}`)}>
```
juliusmarminge, line 18: would have liked a `<Link href="">` here but that seems to apply some nasty styles which i cba fixing so just used `router.push` instead :/

juliusmarminge:
Changed up some other styles while i was at it:

CleanShot 2023-02-08 at 21 17 31

juliusmarminge (approved):

PiotrekPKP:
> Question though, is this the proper way to grab params from dynamic paths?
> 
> Should we not get something from props? I only get a fixed string `[id]` and not the unique id o the "request"
> 
> CleanShot 2023-02-08 at 21 01 33

I think we should use `useSearchParams`: https://expo.github.io/router/docs/features/linking#usesearchparams

juliusmarminge:
> 



> > Question though, is this the proper way to grab params from dynamic paths?
> > Should we not get something from props? I only get a fixed string `[id]` and not the unique id o the "request"
> > CleanShot 2023-02-08 at 21 01 33
> 
> I think we should use `useSearchParams`: https://expo.github.io/router/docs/features/linking#usesearchparams

Oh - feel free to change that in a new PR :)

PiotrekPKP:
> > 
> 
> > > Question though, is this the proper way to grab params from dynamic paths?
> > > Should we not get something from props? I only get a fixed string `[id]` and not the unique id o the "request"
> > > CleanShot 2023-02-08 at 21 01 33
> > 
> > 
> > I think we should use `useSearchParams`: https://expo.github.io/router/docs/features/linking#usesearchparams
> 
> Oh - feel free to change that in a new PR :)

Sure, will look at it :)

juliusmarminge:
I still think it should come from props though :/

PiotrekPKP:
> I still think it should come from props though :/

Why? I think the current approach is ok

juliusmarminge:
> Why? I think the current approach is ok

I've just found that this is so much better in Next.js appdir:

```ts
// app/post/[id]/page.tsx
function Page({ params }: { params: { name: string } }) {
  const { data } = getByName(params.name);
}
```
compared to the old:
```ts
// pages/post/[id].tsx
function Page() {
  const { query } = useRouter();
  if (!query.name) throw new Error(); // <-- notice we have to check this
  const { data } = getByName(query.name);
}
```

PiotrekPKP:
> Member
> ###

I honestly haven't seen that option before, thanks for letting me know it exists haha
Anyways, I'm not sure if it's possible to use with Expo Router...

PR #137, opened 2023-01-31
expo-cli 3.3.0

leifarriens:
Fixes `InvalidEasJsonError: eas.json is not valid.` error when running `eas build` with an eas-cli version < 3.3.0.

`build.preview.ios.resourceClass` is only supported in eas-cli >= 3.3.0

Bildschirmfoto 2023-01-31 um 02 59 05


juliusmarminge (approved):
Thanks 🙏

PR #139, opened 2023-02-02
feat: configuring debugging on vscode

### LINKED ISSUE #112: bug: VSCode Debugging

Create a brand new project from template.

### Describe the bug

Breakpoints are not binding. They bind and work as expected on regular create-t3 (non-monorepo).

### Link to reproduction

 Just set breakpoints in any TRPC router.

### To reproduce

They will not bind in the API routers or anything else.

### Additional information

Awesome work! I'm loving the T3 stack and this Monorepo component. Just having some issues running VSCode debugger and setting breakpoints.

gnllucena:
This PR configures debugging on vscode and fixes #112.

Works for both the api routes, and the turborepo packages.

image

image

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,13 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Next.js",
+      "type": "node-terminal",
+      "request": "launch",
+      "command": "pnpm dev",
+      "cwd": "${workspaceFolder}/apps/nextjs/",
+      "skipFiles": ["<node_internals>/**"]
+    }
+  ]
+}


juliusmarminge (approved):
This looks pretty standard? What was the problem you ran into initially?

gnllucena:
It's very similar to the default launch.json vscode creates, but it has some very important differences for the monorepo architecture.

default launch.json:
```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "node",
            "request": "launch",
            "name": "Launch Program",
            "skipFiles": [
                "<node_internals>/**"
            ],
            "program": "${file}"
        }
    ]
}
```

The default misses the startup parameter we use in create-t3-turbo: `"command": "pnpm dev",`

And most important for the monorepo architecture the `"cwd": "${workspaceFolder}/apps/nextjs/",` parameter.

As we can have multiple different apps, the vscode debugger needs to point to the correct node_modules folder. Without the pointing, the debugger cannot find the dependencies the app needs to startup:

image

PR #157, opened 2023-02-08
fix: packages not using eslint global config

gnllucena:
This PR fixes two issues with `pnpm lint`.

First: the `api`, `auth` and `expo` packages/app are not using the global `.eslintrc.js` config.

Second: the `pnpm lint` script does not continue in the presence of an error in one of the packages/apps, this behavior hides potential lint errors on other packages/apps.

Before:
Pasted Graphic 1

After: 
Pasted Graphic 2

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,12 @@
 /** @type {import("eslint").Linter.Config} */
 const config = {
   extends: ["prettier", "eslint:recommended"],
+  parserOptions: {
+    ecmaVersion: "latest",
+  },
+  env: {
+    es6: true,
+  },
   overrides: [
     {
       extends: [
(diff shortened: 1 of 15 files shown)


juliusmarminge:
> First: the api, auth and expo packages/app are not using the global .eslintrc.js config.

That's incorrect. How eslint configs work is that it'll traverse up the filetree until it finds a config marked `root: true`. You can see that the config is resolved correctly here:

`packages/auth` has override to allow unused vars:
CleanShot 2023-02-08 at 20 17 02

that override removed:
CleanShot 2023-02-08 at 20 17 27

However, I did notice that the eslint CLI wasn't actually running on any files in these packages, so you'd only get the warnings in your editor but the CLI wouldn't report any warnings. I fixed that in [ffe89b3](https://github.com/t3-oss/create-t3-turbo/commit/ffe89b36b23f45d91c142d4ee02587cb335402cf) so now we're running `eslint .` instead of just `eslint` which will make sure these pacakges are linted also with the CLI.

Happy to accept the `--continue` flag though - that seems like a good addition.

juliusmarminge (approved):

PR #171, opened 2023-02-15
Configure EAS Builds to Use M1 Workers

peterkibuchi:
- I configured both `development` and `production` builds to make use of M1 workers; they were previously only configured for `preview` builds. [Expo says](https://blog.expo.dev/m1-workers-on-eas-build-dcaa2c1333ad) that M1 workers generally speed up builds, and will be the default before Q2 2023.
- I indented the `eas.json` file and arranged the json properties alphabetically for better readability.

### DIFF
--- a/apps/expo/eas.json
+++ b/apps/expo/eas.json
@@ -1,15 +1,29 @@
 {
-  "cli": { "version": ">= 3.3.0" },
+  "cli": {
+    "version": ">= 3.3.0"
+  },
   "build": {
     "development": {
       "developmentClient": true,
-      "distribution": "internal"
+      "distribution": "internal",
+      "ios": {
+        "resourceClass": "m1-medium"
+      }
     },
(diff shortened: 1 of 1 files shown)


juliusmarminge:
I believe this should be the default already since we're now on Expo 48?

peterkibuchi:
From what Expo have communicated, yes, it should.

What I'm not entirely sure about is whether we still should include the `"resourceClass": "m1-medium"` in `eas.json`. I initially opened the PR because the M1 workers were explicitly specified for `preview` builds, but not other builds, despite having upgraded to Expo 48. Do you suppose I should edit the PR to remove the lines explicitly specifying to use M1 workers for _all_ of the builds, since they're now supposed to be the default?

juliusmarminge (approved):
whatever - let's be explicit. m1 rocks!

PR #172, opened 2023-02-16
fix: re-enable manypkg in CI

ikapo:
Now that Expo SDK was upgraded to v48 it supports React 18.2, meaning the `manypkg check` can be re-enabled in CI.
In addition, running `manypkg check` resulted in an error so `manypkg fix` sorted the relevant imports.


juliusmarminge (approved):
Thanks :)

PR #175, opened 2023-02-18
fix(renovate): don't run renovate on acme

filipweilid:
This should stop the "Renovate failed to look up the following dependencies" warnings for acme packages

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,5 +2,11 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "config:base"
+  ],
+  "packageRules": [
+    {
+      "matchPackagePatterns": ["^@acme/"],
+      "enabled": false
+    }
   ]
 }


juliusmarminge:
Do you know how we could make it work on acme packages instead? I dont see why we wouldn't want those packages to be bumped

filipweilid:
> Do you know how we could make it work on acme packages instead? I dont see why we wouldn't want those packages to be bumped

Not sure, from what I understand from this https://docs.renovatebot.com/configuration-options/#updateinternaldeps it doesn't seem possible with turborepo

renovate.json
```diff
@@ -2,5 +2,11 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "config:base"
+  ],
+  "packageRules": [
+    {
+      "matchPackagePatterns": ["^@acme/"],
```
juliusmarminge, line 8: This pattern would match @acme/expo and @acme/nextjs too and they are working now

filipweilid: They aren't being used as dependencies anywhere so renovate isn't trying to find them anyway

filipweilid:
Playing around a bit more with this it does seem like renovate supports pnpm workspaces as longs as you use "workspace:*" for the dependencies, renovate will just ignore them as internal dependencies. Setting updateInternalDeps to true didn't make it show up for renovate either.

juliusmarminge:
> Playing around a bit more with this it does seem like renovate supports pnpm workspaces as longs as you use "workspace:*" for the dependencies, renovate will just ignore them as internal dependencies. Setting updateInternalDeps to true didn't make it show up for renovate either.

Hmm - we use renovate at trpc which is a pnpm-workspace monorepo as well... We use the proper versions there though ("@trpc/server": "^10.12.0" for example). Is that the reason?

filipweilid:
> > Playing around a bit more with this it does seem like renovate supports pnpm workspaces as longs as you use "workspace:*" for the dependencies, renovate will just ignore them as internal dependencies. Setting updateInternalDeps to true didn't make it show up for renovate either.
> 
> Hmm - we use renovate at trpc which is a pnpm-workspace monorepo as well... We use the proper versions there though ("@trpc/server": "^10.12.0" for example). Is that the reason?

Doesn't that mean it's not acting as internal dependencies? "@trpc/server" is uploaded to npmjs so renovate is going to find it while acme packages are internal only which is why the warnings are showing up

juliusmarminge:
Oh i think we're talking about different issues then xddd

I mean I want the deps of the iternal packages to be updated (for example when trpc is updated that should be updated across the entire monorepo). Maybe this already works xdddd

filipweilid:
> Oh i think we're talking about different issues then xddd
> 
> I mean I want the deps of the iternal packages to be updated (for example when trpc is updated that should be updated across the entire monorepo). Maybe this already works xdddd

We probably are but I'm still confused xd the internal packages aren't versioned, it's just using * so how is renovate going to update it

juliusmarminge:
> > Oh i think we're talking about different issues then xddd
> > I mean I want the deps of the iternal packages to be updated (for example when trpc is updated that should be updated across the entire monorepo). Maybe this already works xdddd
> 
> We probably are but I'm still confused xd the internal packages aren't versioned, it's just using * so how is renovate going to update it

Oh yea no that's not what I mean. The packages themselves don't need to get updated. I think we're on the same page now xddd

juliusmarminge (approved):

PR #176, opened 2023-02-20
Minor Updates & Configure Expo App Scripts to Use Local Expo CLI

peterkibuchi:
I configured the Expo App Scripts to use the Local Expo CLI instead of the Global Expo CLI. In SDK 46 [Expo introduced a new “Local Expo CLI”](https://medium.com/the-exponent-log/the-new-expo-cli-f4250d8e3421), to replace the “Global Expo CLI” (npm i -g expo-cli). It's lighter, faster, and eliminates the need for devs to install (and keep updating) the Global CLI. See linked article to learn more about the advantages/improvements.

I also made some minor changes to the Expo `app` directory:
1. I destructured the import of `useState` to replace the dot notation implementation.
2. I imported the `SafeAreaView` component from `react-native-safe-area-context` instead of `react-native` because the latter is only available for iOS devices.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,8 +3,9 @@
   "extends": ["config:base"],
   "packageRules": [
     {
-      "matchPackagePatterns": ["^@acme/"],
+      "matchPackagePatterns": ["^@acme/", "eslint-config-acme"],
       "enabled": false
     }
-  ]
+  ],
+  "updateInternalDeps": true
 }


apps/expo/package.json
```diff
@@ -4,9 +4,9 @@
   "main": "index.ts",
   "scripts": {
     "clean": "rm -rf .expo .turbo node_modules",
-    "dev": "expo start --ios",
-    "dev:android": "expo start --android",
-    "dev:ios": "expo start --ios",
+    "dev": "npx expo start --ios",
+    "dev:android": "npx expo start --android",
+    "dev:ios": "npx expo start --ios",
```
juliusmarminge, line 9: Revert these, no need to specify npx here. You only need that when you run it from the terminal

CleanShot 2023-02-23 at 19 59 46@2x

peterkibuchi: Thanks for the correction. I will open a new PR with the `npx` commit excluded.

PR #184, opened 2023-02-23
chore: cleanup config stuff, version bumps

juliusmarminge:


apps/expo/package.json
```diff
@@ -8,26 +8,27 @@
     "dev:android": "expo start --android",
     "dev:ios": "expo start --ios",
     "lint": "eslint .",
-    "type-check": "tsc --noEmit"
+    "lint:fix": "pnpm lint --fix",
+    "FIXME(no-work):type-check": "tsc --noEmit"
```
juliusmarminge, line 12: Started getting a bunch of internal typeerrors from expo-router, despite having `skipLibCheck: true`.

Will hopefully be fixed upstream, not sure what config stuff we can do here?

Errors here: https://app.warp.dev/block/dokPFDcm5lALq8gMiMWnb4

PR #187, opened 2023-02-25
fix: make clean scripts compatible with windows

byCedric:
I just noticed that the clean scripts are using `rm -rf`, which doesn't work on Windows machines. I know, it's probably not the best dev OS, but lots of people are using Windows :)

There are other ways to achieve this, but I found that this is one of the cleanest ways. It does assume that people use git, but even windows users _should_ be using git.

### DIFF
--- a/apps/expo/package.json
+++ b/apps/expo/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "main": "index.ts",
   "scripts": {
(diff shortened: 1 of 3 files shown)


swyftx-simon (approved):

juliusmarminge (approved):
Awesome! Always been scared of git clean cause it wipes .env files and such too, didn't know about these flags :)

PR #192, opened 2023-03-03
refactor: make acme/api devDep in expo

### LINKED ISSUE #164: Does this pattern leak backend into expo app?

n/a

### Describe the bug

Since we're including the full api as a dependency of the RN app, would this not embed all of our api source code into the expo app and ship it to end users? AFAIK, react native does no tree-shaking, so the simple dependency and import inside of RN would cause the source of our api to be pulled in. This would be bad.

I'm not 100% this happens as I cannot find the source code in the final expo bundle, but I can't don't understand why that wouldn't happen. It makes me nervous. Any insight as to why it's safe or not?

### Link to reproduction

https://github.com/t3-oss/create-t3-turbo

### To reproduce

Not sure

### Additional information

_No response_

juliusmarminge:
Closes #164

Makes `@acme/api` a dev dependency in `@acme/expo` since we only need the types and don't want to worry about leaking server code.

Also some other general maintenance work

### DIFF
--- a/apps/expo/app.config.ts
+++ b/apps/expo/app.config.ts
@@ -19,7 +19,7 @@ const defineConfig = (): ExpoConfig => ({
   assetBundlePatterns: ["**/*"],
   ios: {
     supportsTablet: true,
-    bundleIdentifier: "juliusmarminge.expo.com",
+    bundleIdentifier: "your.bundle.identifier",
   },
   android: {
     adaptiveIcon: {
@@ -29,7 +29,7 @@ const defineConfig = (): ExpoConfig => ({
   },
   extra: {
     eas: {
-      projectId: "768478b6-46cd-43a3-904b-f4d5065648d2",
+      projectId: "your-project-id",
     },
   },
   plugins: ["./expo-plugins/with-modify-gradle.js"],
--- a/apps/expo/app.config.ts
+++ b/apps/expo/app.config.ts
@@ -1,6 +1,6 @@
-import { type ConfigContext, type ExpoConfig } from "@expo/config";
+import type { ExpoConfig } from "@expo/config";
 
-const defineConfig = (_ctx: ConfigContext): ExpoConfig => ({
+const defineConfig = (): ExpoConfig => ({
   name: "expo",
   slug: "expo",
   scheme: "expo",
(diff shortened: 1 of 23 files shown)


apps/expo/app.config.ts
```diff
@@ -1,6 +1,6 @@
-import { type ConfigContext, type ExpoConfig } from "@expo/config";
```
juliusmarminge, line 1: EAS didnt like this import for whatever reason 🤷‍♂️

apps/expo/package.json
```diff
@@ -30,9 +28,12 @@
     "react-dom": "18.2.0",
     "react-native": "0.71.3",
     "react-native-safe-area-context": "4.5.0",
-    "react-native-screens": "~3.20.0"
+    "react-native-screens": "~3.20.0",
+    "superjson": "1.9.1"
   },
   "devDependencies": {
+    "@acme/api": "*",
```
juliusmarminge, line 35: superjson can no longer be importerad from the api package as we dont share any runtime code from it

apps/nextjs/package.json
```diff
@@ -3,41 +3,41 @@
...
-    "build": "pnpm with-env next build",
+    "build": "next build",
     "clean": "rm -rf .next .turbo node_modules",
-    "dev": "pnpm with-env next dev",
+    "dev": "next dev",
     "lint": "SKIP_ENV_VALIDATION=1 next lint",
     "lint:fix": "pnpm lint --fix",
     "start": "next start",
-    "type-check": "tsc --noEmit",
-    "with-env": "dotenv -e ../../.env --"
```
juliusmarminge, line 13: Moved to root

apps/nextjs/src/utils/api.ts
```diff
@@ -30,14 +30,4 @@ export const api = createTRPCNext<AppRouter>({
...
- * Inference helpers for input types
- * @example type HelloInput = RouterInputs['example']['hello']
- **/
-export type RouterInputs = inferRouterInputs<AppRouter>;
-
-/**
- * Inference helpers for output types
- * @example type HelloOutput = RouterOutputs['example']['hello']
- **/
-export type RouterOutputs = inferRouterOutputs<AppRouter>;
```
juliusmarminge, line 43: Moved to the api packade to reduce DRY

PR #199, opened 2023-03-08
refactor: move expo router into src

juliusmarminge:
What do you think @PiotrekPKP, worth the hacky workaround?


PiotrekPKP:
> What do you think @PiotrekPKP, worth the hacky workaround?

I think it's worth it! Will test it out today and let you know to make sure everything's right

PiotrekPKP:
@juliusmarminge Everything looks nice and I don't think this hacky workaround will annoy anyone so yeah, why not!

Anyway, I found a fix for the TS error that ws caused by the `require.context` call (can't push to this PR so here's the link): https://github.com/PiotrekPKP/create-t3-turbo/commits/expo-srcapp

PR #200, opened 2023-03-08
refactor: move expo router into src (no ts error)

PiotrekPKP:
Same as #199 but with fixed TS error


PiotrekPKP:
> Does it still build on EAS?

Looks like it does :)
image

juliusmarminge (approved):
Nice

PR #215, opened 2023-03-21
Use pnpm to manage node version

kevgk:
With this change, there is no need for nvm, since pnpm can manage node version too.


juliusmarminge:
Does this work for other node version tools? Will they automatically detect this and change node version if i `cd` into the project?

With .nvmrc, your node-version is automatically switched when `cd create-t3-turbo` - this also works for `fnm` and I think other node version tools as well.

kevgk:
I assumed that pnpm is the chosen package manager for this repo. I don't think other version managers support this format. But is that really a requirement?

juliusmarminge:
> I assumed that pnpm is the chosen package manager for this repo. I don't think other version managers support this format. But is that really a requirement?

Can PNPM select the node version I'm running? Didn't know that

PR #216, opened 2023-03-21
Use workspace prefix for workspace dependencies

kevgk:
Use the workspace prefix in package.json, for workspace dependencies, to prevent installation of unwanted packages.

### DIFF
--- a/apps/expo/package.json
+++ b/apps/expo/package.json
@@ -32,8 +32,8 @@
     "superjson": "1.9.1"
   },
   "devDependencies": {
-    "@acme/api": "*",
-    "@acme/tailwind-config": "*",
+    "@acme/api": "workspace:*",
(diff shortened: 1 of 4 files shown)


juliusmarminge:
Manypkg doesn't support this. I think we could use this though: https://pnpm.io/npmrc#prefer-workspace-packages

kevgk:
I guess that would be the next best thing.

odylomv:
I found that setting prefer-workspace-packages to true and save-workspace-protocol to false achieves this goal and stops pnpm from constantly trying to add the "workspace:" prefix (for example on pnpm update)

juliusmarminge:
I dont want to sacrifise manypkg as i think it adds a ton of value.

PRs for any other solution than `workspace:*` is welcome 🙏

PR #240, opened 2023-03-26
Update package.json

Diogovski:
When you run pnpm lint, 'SKIP_ENV_VALIDATION' will throw a 'not recognized command' error, adding SET on the script of lint inside nextjs app package.json solves it.
Screenshot_2
Screenshot_3

### DIFF
--- a/apps/nextjs/package.json
+++ b/apps/nextjs/package.json
@@ -6,7 +6,7 @@
     "build": "next build",
     "clean": "git clean -xdf .next .turbo node_modules",
     "dev": "next dev",
-    "lint": "SKIP_ENV_VALIDATION=1 next lint",
+    "lint": "SET SKIP_ENV_VALIDATION=1 next lint",
     "lint:fix": "pnpm lint --fix",
     "start": "next start",
     "type-check": "tsc --noEmit"


odylomv:
Both yours and the original command are platform-specific solutions. Your command doesn't work on Linux and the original does not work on Windows. I guess dotenv-cli should be used like in the root package.json.

odylomv:
We can use this opportunity to solve another issue at the same time. Currently, running "pnpm --filter nextjs dev", as instructed in the README to run nextjs seperately, does not work because of env variables. Maybe we can add a dotenv script in each app's package.json to make them work individually and also solve this by running "dotenv -v SKIP_ENV_VALIDATION=1 next lint"

juliusmarminge:
> We can use this opportunity to solve another issue at the same time. Currently, running "pnpm --filter nextjs dev", as instructed in the README to run nextjs seperately, does not work because of env variables. Maybe we can add a dotenv script in each app's package.json to make them work individually and also solve this by running "dotenv -v SKIP_ENV_VALIDATION=1 next lint"

Yea I talked to Anthony from Turbo about this and they are cooking up something to fix our needs here. I had dotenv in the apps before but moved cause their docs suggested you keep it in root... Feel free to open a PR switching it back

PR #242, opened 2023-03-27
Prevent pnpm from adding the workspace: prefix while preferring local packages

odylomv:
After previous conversation on #216 , I have made the necessary changes to prevent pnpm from trying to add the "workspace:" prefix on local packages, while still preferring local packages. I have also modified the local package versions on all package.json files to be consistent with each other.

### DIFF
--- a/apps/expo/package.json
+++ b/apps/expo/package.json
@@ -32,9 +32,9 @@
     "superjson": "1.9.1"
   },
   "devDependencies": {
-    "@acme/api": "*",
-    "@acme/eslint-config": "*",
-    "@acme/tailwind-config": "*",
+    "@acme/api": "^0.1.0",
+    "@acme/eslint-config": "^0.1.0",
+    "@acme/tailwind-config": "^0.1.0",
     "@babel/core": "^7.20.0",
     "@babel/preset-env": "^7.20.0",
     "@babel/runtime": "^7.20.0",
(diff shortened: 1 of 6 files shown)


apps/expo/package.json
```diff
@@ -32,9 +32,9 @@
     "superjson": "1.9.1"
   },
   "devDependencies": {
-    "@acme/api": "*",
-    "@acme/eslint-config": "*",
-    "@acme/tailwind-config": "*",
+    "@acme/api": "^0.1.0",
+    "@acme/eslint-config": "^0.1.0",
+    "@acme/tailwind-config": "^0.1.0",
```
juliusmarminge, line 37: what's wrong with "*" in your opinion? to me it quickly "clicks" that this is an internal package since it's not really versioned

juliusmarminge (commented):
Can you resolve conflicts please?

Also not sure about the explicit versions on the internal packages, is it something pnpm enforces or personal preference?

apps/expo/package.json
odylomv: pnpm always changes it back to "^0.1.0" in my repos, I guess it also happened on the root package.json when you bumped deps.

odylomv:
I resolved the conflicts. I'm not sure if the way I did it the cleanest, lmk if u need something.

PR #247, opened 2023-03-28
Move dotenv-cli from root to necessary packages

odylomv:
After previous discussion in #240 , this PR moves dotenv-cli to each package that requires environment variables. 

This solves the issue of individual package scripts failing due to env viariables missing (for example "pnpm --filter nextjs dev"). As a side-effect, this PR includes a platform-agnostic solution to the nextjs lint command. 

I removed the dotenv-cli package from the root package.json as it is no longer used.


juliusmarminge (approved):
I updated to pnpm@8 so the lockfile is new. Can you resolve the conflicts please 🙏🏼

odylomv:
Conflicts resolved.

PR #255, opened 2023-03-31
feat: implemented docker support for nextjs app

EkkoKo:


apps/nextjs/next.config.mjs
```diff
@@ -2,10 +2,12 @@
  * Run `build` or `dev` with `SKIP_ENV_VALIDATION` to skip env validation.
  * This is especially useful for Docker builds and Linting.
  */
-// !process.env.SKIP_ENV_VALIDATION && (await import("./src/env.mjs"));
+!process.env.SKIP_ENV_VALIDATION && (await import("./src/env.mjs"));
 
 /** @type {import("next").NextConfig} */
 const config = {
+  /** For docker build, more information: https://nextjs.org/docs/advanced-features/output-file-tracing */
+  output: "standalone",
```
juliusmarminge, line 10: put this as part of the README docs instead. i dont want this to be the default resort

apps/nextjs/package.json
```diff
@@ -27,18 +27,17 @@
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "superjson": "1.9.1",
+    "@types/node": "^18.15.11",
+    "@types/react": "^18.0.31",
+    "tailwindcss": "^3.3.1",
+    "postcss": "^8.4.21",
+    "autoprefixer": "^10.4.14",
```
juliusmarminge, line 34: im no docker expert but why do these need to be production deps all of a sudden?

packages/db/package.json
```diff
@@ -12,11 +12,11 @@
     "with-env": "dotenv -e ../../.env --"
   },
   "dependencies": {
+    "prisma": "^4.12.0",
```
juliusmarminge, line 15: same here - `prisma` is a devDep, why does it need to be a prod dep now?

juliusmarminge (changes requested):
Haven't tested it out myself yet but got some initial thoughts

apps/nextjs/package.json
EkkoKo: when you run `pnpm install --prod `it only installs `dependencies` and when nextjs is trying to create to build it fails because it looks for those dependencies.

packages/db/package.json
EkkoKo: in the turbo.json we declare that before build it should run `db:generate`, `db:generate` runs `prisma` and because we only install `dependencies` I had to add it.

ChristianKuri:
I havent tested it, but im looking forward using it. I think its also a good idea to include a docker compose file.

EkkoKo:
> I havent tested it, but im looking forward using it. I think its also a good idea to include a docker compose file.

I have included a `docker-compose` file, you should definitely check it out :) 
It is also mentioned it in the `README`

hichemfantar:
Any plans to merge this?
What's blocking this from getting approved.

apps/nextjs/package.json
albertilagan: imo just remove `--prod`, revert to `output: "standalone"`.

packages/db/package.json
albertilagan: As long as you're using `output: "standalone"` devDep will not be included on the final image, so imo you don't need `--prod` when running `pnpm install`, specially since you're doing `prune` anyway.

akutruff:
I've burned all day trying to get docker setup and it looks like everything out there is either broken or results in an image size of a GB in size.  Anyone have a working docker setup that doesn't alter dependencies?

albertilagan:
> I've burned all day trying to get docker setup and it looks like everything out there is either broken or results in an image size of a GB in size. Anyone have a working docker setup that doesn't alter dependencies?

Here,
```Dockerfile
ARG NODE_VERSION=18.17.1
ARG ALPINE_VERSION=3.17

FROM node:${NODE_VERSION}-alpine${ALPINE_VERSION} AS base

RUN npm i -g pnpm@8.7.5
RUN npm i -g turbo@1.10.14

FROM base AS builder

# Set working directory
WORKDIR /app
COPY . .
RUN turbo prune --scope=test-app --docker

# Add lockfile and package.json's of isolated subworkspace
FROM base AS installer
WORKDIR /app

ARG TURBO_TEAM
ARG TURBO_TOKEN

ENV TURBO_TEAM=$TURBO_TEAM
ENV TURBO_TOKEN=$TURBO_TOKEN

# First install the dependencies (as they change less often)
COPY .gitignore .gitignore
COPY --from=builder /app/out/json/ ./
COPY --from=builder /app/out/pnpm-lock.yaml ./pnpm-lock.yaml
RUN pnpm install --frozen-lockfile

# Build the project
COPY --from=builder /app/out/full/ ./
RUN CI=true SKIP_ENV_VALIDATION=1 pnpm build --filter=test-app

FROM node:${NODE_VERSION}-alpine${ALPINE_VERSION} AS runner
WORKDIR /app

COPY --from=installer /app/apps/test-app/next.config.mjs .
COPY --from=installer /app/apps/test-app/package.json .

# Automatically leverage output traces to reduce image size
# https://nextjs.org/docs/advanced-features/output-file-tracing
COPY --from=installer /app/apps/test-app/.next/standalone ./
COPY --from=installer /app/apps/test-app/.next/static ./apps/test-app/.next/static
COPY --from=installer /app/apps/test-app/public ./apps/test-app/public

CMD node apps/test-app/server.js
```

albertilagan:
> @albertilagan Thanks for the reply! A couple of things I found that you may want to add:
> 
> * Install `libc6-compat` to alpine base image. (see comment inline)
> * Can copy `turbo.json` and use `turbo run build` with ellipses to do the build with dependencies.
> * Added a commented out line to copy an `.env` file in case the build is not being used with `docker compose` Full example in the next.js repo [here](https://github.com/vercel/next.js/blob/305bb015060e82be3e6ae59a3d063c11a3e207f9/examples/with-docker-multi-env/docker/production/Dockerfile)
> 
> ```dockerfile
> FROM node:18-alpine AS base
> # Check https://github.com/nodejs/docker-node/tree/b4117f9333da4138b03a546ec926ef50a31506c3#nodealpine to understand why libc6-compat might be needed.
> RUN apk add --no-cache libc6-compat
> 
> FROM base AS builder
> WORKDIR /app
> 
> RUN npm install -g turbo
> COPY . .
> RUN turbo prune --scope=@acme/nextjs --docker
> 
> FROM base as installer
> WORKDIR /app
> 
> RUN npm install -g pnpm
> RUN npm install -g turbo
> 
> ARG TURBO_TEAM
> ARG TURBO_TOKEN
> 
> ENV TURBO_TEAM=$TURBO_TEAM
> ENV TURBO_TOKEN=$TURBO_TOKEN
> 
> # First install dependencies (as they change less often)
> COPY .gitignore .gitignore
> COPY --from=builder /app/out/json/ .
> COPY --from=builder /app/out/pnpm-lock.yaml ./pnpm-lock.yaml
> RUN pnpm install --frozen-lockfile
> 
> # Build the project and its dependencies
> COPY --from=builder /app/out/full/ .
> COPY turbo.json turbo.json
> 
> ## This would be useful for browser environment variables that are actually baked at build time and you aren't passing them in otherwise.
> # COPY .env.production .env.production  
> RUN CI=true SKIP_ENV_VALIDATION=true turbo run build --filter=@acme/nextjs...
> 
> FROM base AS runner
> WORKDIR /app
> 
> EXPOSE 3000
> ENV PORT 3000
> 
> ENV NODE_ENV production
> ENV HOSTNAME localhost
> 
> # Don't run production as root
> RUN addgroup --system --gid 1001 nodejs
> RUN adduser --system --uid 1001 nextjs
> USER nextjs
> 
> COPY --from=installer /app/apps/nextjs/next.config.mjs .
> COPY --from=installer /app/apps/nextjs/package.json .
> 
> # Automatically leverage output traces to reduce image size
> # https://nextjs.org/docs/advanced-features/output-file-tracing
> COPY --from=installer --chown=nextjs:nodejs /app/apps/nextjs/.next/standalone ./
> COPY --from=installer --chown=nextjs:nodejs /app/apps/nextjs/.next/static ./apps/nextjs/.next/static
> COPY --from=installer --chown=nextjs:nodejs /app/apps/nextjs/public ./apps/nextjs/public
> 
> CMD node apps/nextjs/server.js
> ```
> 
> For anyone reading this out of context, you also want a .dockerignore in your root
> 
> ```gitignore
> node_modules
> 
> # This line is **EXTREMELY** important
> # https://github.com/vercel/turbo/issues/1997#issuecomment-1273565773
> **/node_modules
> ```
> 
> You also want standalone as @albertilagan pointed out. Here's a full file that supports multi package:
> 
> ```js
> import "./src/env.mjs";
> import "@acme/auth/env.mjs";
> 
> /** @type {import("next").NextConfig} */
> const config = {
>   reactStrictMode: true,
>   /** Enables hot reloading for local packages without a build step */
>   transpilePackages: ["@acme/api", "@acme/auth", "@acme/db"],
>   /** We already do linting and typechecking as separate tasks in CI */
>   eslint: { ignoreDuringBuilds: true },
>   typescript: { ignoreBuildErrors: true },
>   output: 'standalone',
> };
> 
> export default config;
> ```

You're welcome :) Yeah feel free to extend it to whatever you need.
we re-use image for different environment hence we're not adding `.env` files on build time,
      - server env are injected when running the image
      - browser env are injected on runtime as well using docker `ENTRYPOINT` with custom env substitution script that modifies the bundle output.

adamspotlite:
@albertilagan great work! are you checking the `env` at all before the injection? We accidentally shipped an image that ultimately didn't have the proper env vars, but it wasn't picked up until it was already serving traffic..

any chance you know how to check the `env`  before the build to validate it? I currently have the docker setup working and deploying properly but would love to add a pre-build step to ensure that the env that I _will_ be applying to the container will be valid. Any ideas?

apps/nextjs/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+ENV NODE_ENV production
+ENV CI         true
+ENV SKIP_ENV_VALIDATION true
+
+COPY .gitignore .gitignore 
+COPY  --from=builder /app/tsconfig.json ./tsconfig.json
+COPY --from=builder /app/out/json .
+COPY --from=builder /app/out/pnpm-lock.yaml\* ./
+
+RUN yarn global add pnpm && pnpm fetch --prod && pnpm install -r --offline --prod
```
Sivanwol, line 23: why mix with yarn and pnpm ??

PR #258, opened 2023-04-01
feat: add prisma studio command

austin-karren:
It would be great if there was a convenient way to run prisma studio with `pnpm db:studio` similar to the `pnpm db:push` command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "clean:workspaces": "turbo clean",
     "db:generate": "turbo db:generate",
     "db:push": "turbo db:push db:generate",
+    "db:studio": "turbo db:studio",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -11,6 +11,7 @@
     "clean:workspaces": "turbo clean",
     "db:generate": "turbo db:generate",
     "db:push": "turbo db:push db:generate",
+    "db:studio": "turbo db:studio",
```
juliusmarminge, line 14: no reason to run this with turbo right? no caching, no task dep.

I'd just run it with a `pnpm --filter db db:studio` or something

austin-karren: That's a great point actually

Issue #185, opened 2023-02-23
fix: expo `type-check` script disabled


juliusmarminge:
Started getting a bunch of internal typeerrors from expo-router, despite having `skipLibCheck: true`.

Will hopefully be fixed upstream, not sure what config stuff we can do here?

Errors here: https://app.warp.dev/block/dokPFDcm5lALq8gMiMWnb4

_Originally posted by @juliusmarminge in https://github.com/t3-oss/create-t3-turbo/pull/184#discussion_r1116239647_

PiotrekPKP:
I don't think it's an issue anymore? At least it's working for me on a clean repo clone.
![image](https://user-images.githubusercontent.com/53144702/229302826-73cec123-b673-4f63-9d65-4989c48c6a64.png)

FelipeEmos:
Just in case there's someone else struggling with this and wants to fix it by hand, the problem was fixed after updating the `@types/react` library.

I found that by using git bisect and the commit that fixed it was: https://github.com/t3-oss/create-t3-turbo/commit/7c9361acfc2bef1202b4faf9c1d33b1c53d06eec

PR #259, opened 2023-04-01
fix: bring back `type-check` script

PiotrekPKP:
Brings back the `type-check` script as the expo router issue got fixed (see https://github.com/t3-oss/create-t3-turbo/issues/185#issuecomment-1493038578)


juliusmarminge (approved):
Thanks

PR #260, opened 2023-04-01
fix: change babel's config type

PiotrekPKP:
Changes the config type and replaces `api.cache(true)` with `api.cache.forever()`. Docs [say it's the same](https://babeljs.io/docs/config-files#apicache) and the type for `api.cache(true)` was wrong for whatever reason.


juliusmarminge (approved):
:)

PR #270, opened 2023-04-07
feat(expo): add import alias & fix relative paths

FelipeEmos:
Specially for big projects it's not nice having to do relative imports all the time. 🙃 

- This PR removes relative imports from the `expo` app
- This PR gives the `expo` app an import alias equivalent to the `nextjs` app, the `~/`
    - Which makes them more similar


apps/expo/tsconfig.json
```diff
@@ -1,6 +1,10 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "~/*": ["../../apps/expo/src/*"]
```
juliusmarminge, line 6: Why ../../app/expo ?

I know i had issues getting this to work when i added the nextjs aliases. Have you tried making a build after adding these aliases?

FelipeEmos: @juliusmarminge, you are right twice, thanks for pointing those out to me. I don't know what was in my mind.

1. There's no need to go back and forward with the path of course... I guess something bad happened with my types resolution and when it worked I stopped reasoning.
2. I forgot to configure the bundler alias. Now this PR has 1 new commit which configures it and I successfully built the expo app with this new config.

juliusmarminge (approved):

juliusmarminge (approved):

PR #300, opened 2023-04-20
fix localhost determination during expo local dev

johnslemmer:
Turns out when just using expo go and local development what was originally here works fine.  But if you end up building your own dev client (https://docs.expo.dev/develop/development-builds/introduction/#what-is-an-expo-dev-client) then `manifest` will no longer be around. Instead you need to use `manifest2`.

See description of `manifest` and `manifest2` here: https://docs.expo.dev/versions/latest/sdk/constants/#nativeconstants

image


juliusmarminge (approved):
Nice! Didn't know about this

johnslemmer:
@juliusmarminge not sure of your process here.  But I think I need you to merge, or authorize Vercel or both.

juliusmarminge:
Ah sorry bout that

PR #307, opened 2023-04-21
Simplify `as unknown as ...` cast in db/index.ts

mmkal:
Switch to a slightly more "honest" cast of `globalThis` in the db index file.

`globalThis` _is_ an object which has an optional `prisma` property - it doesn't make much difference in this case but casting to `as unknown` allows making a slightly more unsafe cast, which isn't needed here - at least I don't think it is.

### DIFF
--- a/packages/db/index.ts
+++ b/packages/db/index.ts
@@ -2,7 +2,7 @@ import { PrismaClient } from "@prisma/client";
 
 export * from "@prisma/client";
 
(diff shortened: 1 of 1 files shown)


juliusmarminge (approved):
Thanks :) not sure why i had the unknown if it wasn't necessary. Iirc I got errors without it before 🤷‍♂️

PR #323, opened 2023-04-27
feat: t3-env separation

### LINKED ISSUE #87: feat: better way to validate / import env vars

juliusmarminge:
Closes #87


bhatvikrant:
When is this getting merged? @juliusmarminge

melkir:
Can confirm, this works well 👍

PR #326, opened 2023-04-28
Make an @acme/env package usable in the other monorepo packages

johnslemmer:
After @juliusmarminge just pushed the transition to https://env.t3.gg/ I saw an opportunity.  This PR creates a simple @acme/env package that can be installed in the other monorepo packages so you can more systematically and typesafely access environment variables.

The hack that makes this work is that the package's main export is `index.mjs` however `index.ts` sym-links to it in order to provide good typing but still be usable inside of `next.config.mjs`.

I already did this in my own project and it works great.  Thanks for the inspiration @juliusmarminge and the work you did on t3-env!  I did see you were [playing around](https://github.com/t3-oss/create-t3-turbo/pull/135) with something for ENVs a few months ago.  But this seemed so simple in comparison and hopefully an easy win.

### DIFF
--- a/apps/nextjs/next.config.mjs
+++ b/apps/nextjs/next.config.mjs
@@ -2,7 +2,7 @@
  * Run `build` or `dev` with `SKIP_ENV_VALIDATION` to skip env validation.
  * This is especially useful for Docker builds and Linting.
  */
-// !process.env.SKIP_ENV_VALIDATION && (await import("./src/env.mjs"));
+await import("@acme/env");
 
 /** @type {import("next").NextConfig} */
 const config = {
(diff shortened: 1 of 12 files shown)


juliusmarminge:
Problem with a shared package is it will bust the cache unnecessarily, as I stated here: https://github.com/t3-oss/create-t3-turbo/pull/135#issuecomment-1414339096

I started in #323 with a bit of a different approach where each package/app validates their own envs which means a cache bust in one place shouldn't invalidate everything

johnslemmer:
I'll keep an eye on my builds/CI the next few days and get back to you. Also, I'm curious with how little environment variables change how much this will be an issue. Maybe worth the rare cache bust for type-safe ENV references 🤷‍♂️. However, your individualized ENV validation for each package seems like the best approach.

johnslemmer:
@juliusmarminge saw how you addressed this in your commit today. I like it. Any reason DATABASE_URL shouldn't be pulled in to an env.mjs inside @acme/db?

PR #359, opened 2023-05-14
Fix/nativewind ts config

cjkihl:
# Tailwind typescript config file for expo/nativewind.

This PR allows expo and nativewind to use `tailwind.config.ts` configuration just like the NextJS app.

## How to test

1. Run `pnpm install`
2. Start the emulator and run `pnpm dev` 
3. Make sure babel compiles and tailwind configuration is applied.

4. (Optional) You might need to clear the cache
https://docs.expo.dev/troubleshooting/clear-cache-macos-linux/
https://docs.expo.dev/troubleshooting/clear-cache-windows/


apps/expo/babel.config.js
```diff
@@ -9,7 +31,12 @@ module.exports = function (api) {
   return {
     presets: ["babel-preset-expo"],
     plugins: [
-      "nativewind/babel",
+      [
+        "nativewind/babel",
+        {
+          tailwindConfig: lazyTranspileTailwindConfig(),
```
juliusmarminge, line 37: Can we use the `loadConfig` exported from `tailwindcss/loadConfig`? It uses jiti and sucrase too.

Would feel more "official"

cjkihl: Good idea, fewer dependencies as well :)

PR is updated with `loadConfig` from Tailwind.

apps/expo/package.json
```diff
@@ -12,6 +12,7 @@
     "type-check": "tsc --noEmit"
   },
   "dependencies": {
+    "@acme/tailwind-config": "^0.1.0",
```
juliusmarminge, line 15: Required?

cjkihl: Good point, :) 
It shouldn't be since it's only for the build, but I got typescript errors earlier and I saw that it was a `dependency` in `@acme/nextjs` as well so I thought there was something I don't fully understand yet.

However when I ran `pnpm run clean`, installed everything and reset TS server I realize it was a false TS error and everything runs fine with `@acme/tailwind-config` as a `devDependency`.

PR updated

PR #371, opened 2023-05-20
Minor Changes to polish up the Repo

peterkibuchi:
This is a simple polish commit. It does not fix any bugs or add any features; it just makes the repo easier and more pleasant to work with.

Changes:
- Re-ordered recommended extensions to appear in alphabetical order
- Configured Prisma Extension as default formatter for Prisma `.schema` files
- Added some spaces where code was congested to make it more readable
- Added `eslint` and `tailwind` config packages to README.md


.vscode/settings.json
```diff
@@ -1,15 +1,14 @@
...
+  "editor.codeActionsOnSave": { "source.fixAll.eslint": true },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   "eslint.rules.customizations": [{ "rule": "*", "severity": "warn" }],
   "eslint.workingDirectories": [
     { "pattern": "apps/*/" },
     { "pattern": "packages/*/" }
   ],
-  "typescript.tsdk": "node_modules/typescript/lib",
+  "[prisma]": { "editor.defaultFormatter": "Prisma.prisma" },
```
juliusmarminge, line 10: Technically this should go first if were doing alphabetical order

packages/api/src/router/index.ts
juliusmarminge, line 1: This seems like an over-abstraction and just leads to more work?

juliusmarminge (changes requested):

.vscode/settings.json
peterkibuchi: Yeah, nice spot; hadn't thought of that.

packages/api/src/router/index.ts
peterkibuchi: I had a debate with myself on this as well; is it worth the extra file/work? I figured that in the event someone uses this to build a large, complex application with (many) more routers, having a line for each import (in `root.ts`) would muddy the code, make it less elegant/readable. In the event they don't have as many routers, they wouldn't have to touch the `index.ts` file much anyway, and the code would still be (a touch) more elegant.

I don't feel it makes that much of a difference in amount of work/readability introduced either way, and it may be a matter of personal preference in the end. But if the `index.ts` pattern does nothing else, it provides an example on how to implement it elsewhere where it may be more neccessary, e.g. a components folder.

_It's your call_ 👀

juliusmarminge (approved):

PR #386, opened 2023-06-02
feat: add solito

jmcmullen:
This PR adds solito to the project.


jmcmullen:
My bad, didn't realise this was created upstream.

PR #405, opened 2023-06-14
chore: Replace React.FC with basic function syntax

BrightonMboya:
Remove React.FC usage in templates, replacing with standard function syntax as per Cje tweet https://twitter.com/ccccjjjjeeee/status/1668694515484336145

### DIFF
--- a/apps/expo/src/app/index.tsx
+++ b/apps/expo/src/app/index.tsx
@@ -6,30 +6,30 @@ import { FlashList } from "@shopify/flash-list";
 
 import { api, type RouterOutputs } from "~/utils/api";
 
-const PostCard: React.FC<{
+function PostCard(props: {
   post: RouterOutputs["post"]["all"][number];
   onDelete: () => void;
-}> = ({ post, onDelete }) => {
+}) {
   const router = useRouter();
 
   return (
     <View className="flex flex-row rounded-lg bg-white/10 p-4">
       <View className="flex-grow">
-        <TouchableOpacity onPress={() => router.push(`/post/${post.id}`)}>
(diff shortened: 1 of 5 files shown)


apps/expo/src/app/index.tsx
```diff
@@ -118,7 +118,9 @@ const Index = () => {
           data={postQuery.data}
           estimatedItemSize={20}
           ItemSeparatorComponent={() => <View className="h-2" />}
-          renderItem={(p) => (
+          renderItem={(p: {
+            item: { id: string; content: string; title: string };
```
juliusmarminge, line 122: This should be inferred

apps/nextjs/src/pages/index.tsx
```diff
@@ -110,15 +110,17 @@ const Home: NextPage = () => {
...
-                      return (
-                        <PostCard
-                          key={p.id}
-                          post={p}
-                          onPostDelete={() => deletePostMutation.mutate(p.id)}
-                        />
-                      );
-                    })}
+                    {postQuery.data?.map(
+                      (p: { id: string; title: string; content: string }) => {
```
juliusmarminge, line 114: Again, should be inferred.

Did yii ok u forget to run db:generate? Might have to restart the TS server afterwards too

juliusmarminge (changes requested):
Remove the explicit typings of the `p` variables

apps/nextjs/src/pages/index.tsx
BrightonMboya: I think I didn't restart the TS server thats why I kept on seeing the squiggly lines. I have removed it rn

juliusmarminge (approved):

PR #424, opened 2023-07-15
feat: use nextjs appdir (client components)

juliusmarminge:
Uses appdir

- (nested) layouts
- new nextauth, edge compat if you swap out prisma or change to jwt's (https://auth-docs-git-feat-nextjs-auth-authjs.vercel.app/guides/upgrade-to-v5#edge-compatibility)
- new react-query streaming to support `<Suspense />` whilst keeping all the data fetching in client components (for now, might add a server link later)


nikospapcom:
@juliusmarminge I want to ask you why you are using `@auth/core` package? I think it's not stable and has a lot of problems
e.g. in `signIn` callback, the `user` param returns undefined when user don't exist in the database, as a result it's really hard to use it :) 

Any suggestion about it?

PR #435, opened 2023-07-18
Bump expo to 49.0.0

dnishiyama:
I was able to get expo 49 running on an M1 Mac. 

While running on `main` I ran into this bug: https://github.com/postcss/postcss/issues/412. add using `"tailwindcss": "3.3.2"` alone resolved it. But that happened before this work, so I've left it off the PR.


apps/expo/app.config.ts
```diff
@@ -32,7 +32,7 @@ const defineConfig = (): ExpoConfig => ({
       // projectId: "your-project-id",
     },
   },
-  plugins: ["./expo-plugins/with-modify-gradle.js"],
+  plugins: ["expo-router", "./expo-plugins/with-modify-gradle.js"],
```
juliusmarminge, line 35: what does this do?

i don't see this mentioned in any docs + we already have it in the babel.config

dnishiyama: Well that is interesting! I thought I followed [the guide](https://blog.expo.dev/introducing-expo-router-v2-3850fd5c3ca1) exactly, but now I can't find it. I guess I misread it. I just ran it without that line and it was fine. So I'll push a change that drops it.

juliusmarminge:
So in #429 I did some other stuff and cleaned up some of the hacks we had for Expo 48, so I'm going to go ahead with that one but thanks for clearing out the PostCSS issue!

dnishiyama:
Sounds good! Glad I could help

PR #446, opened 2023-07-21
fix: update .nvmrc to use latest lts node 18 version

jmcmullen:
The package.json contains:
```
"engines": {
  "node": ">=v18.17.0"
},
```

But the .nvmrc file requires any version of node 18. This means if someone has a version of Node 18 but lower than 18.17.0, the yarn install will fail. Resulting in the following error when creating a project:

```
>   Installing dependencies...
>>> yarn install has failed.
```

The following change will tell Node Version Manager to use the latest release of node 18.


.nvmrc
```diff
@@ -1 +1 @@
-18
\ No newline at end of file
+lts/hydrogen
```
juliusmarminge, line 1: Why not 18.17 just for consistency. Might work better with renovate too?

jmcmullen: Sure thing, updated. I originally picked ts/hydrogen so the file doesn't need to be updated with every node 18 release.

PR #469, opened 2023-08-03
feat: Credentials Authentication

tyrauber:
This PR adds email / password authentication to create-t3-turbo using the NextAuth Credentials Provider.

Although NextJS discourages password authentication, for companies with an existing user base with existing credentials, it is difficult to not support password authentication. NextAuth also attempts to disable database sessions if using the Credentials Provider.

This PR demonstrates how to use the NextAuth Credential Provider to create JWT tokens **and** Database sessions.

The `authorize` method validates the email, hashes the password with bcrypt, compares password hashes, creates a database session, and returns the user object and sessionToken.

The `jwt` callback is used to add the sessionToken to the token object used to generate the JWT, tying the JWT to the database session, and validates the current JWT token against the database.

A `signOut` event is used delete the session from the database if the sessionToken is present in the JWT.

This should work with additional providers such as discord.

### DIFF
--- a/apps/nextjs/src/app/page.tsx
+++ b/apps/nextjs/src/app/page.tsx
@@ -28,12 +28,20 @@ async function AuthShowcase() {
 
   if (!session) {
     return (
-      <SignIn
-        provider="discord"
-        className="rounded-full bg-white/10 px-10 py-3 font-semibold text-white no-underline transition hover:bg-white/20"
-      >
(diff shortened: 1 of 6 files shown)


juliusmarminge:
Sorry for the late response, but we don't wanna provide a credentials provider by default. You should know what pitfalls there are when adding password auth and providing a default isn't our responsibility.

PR #528, opened 2023-08-18
fix: Remove `SafeAreaProvider`

tomheaton:
`SafeAreaProvider` is automatically added by `expo-router`.


juliusmarminge (approved):
Thanks, didn't know they wrapped it by default. Is the package a peerDep of `expo-router` so we still need it installed, or can we remove that aswell?

tomheaton:
Yeah its a peerDep so still needed on our end.
And yeah its pretty hidden in the documentation [here](https://docs.expo.dev/router/migrate/from-react-navigation/#migration).

PR #535, opened 2023-08-21
fix(api): Make delete post endpoint protected

solofeed:
I think the delete post endpoint should be protected. 
It would also serve as a better example for the example.


juliusmarminge (approved):
Yes.

Part of the reason they were public before was to make them work from the expo app, but I guess it's better for the sake of showing good practises to have it private

Issue #511, opened 2023-08-16
bug: drizzle adapter with postgres ends with `error: relation "t3turbo_account" does not exist`


BleedingDev:
### Provide environment information

System:
    OS: Windows 10 10.0.25300
    CPU: (16) x64 AMD Ryzen 7 6800HS with Radeon Graphics
    Memory: 12.63 GB / 31.26 GB
  Binaries:
    Node: 20.5.1 - ~\AppData\Local\Volta\tools\image\node\20.5.1\node.EXE
    Yarn: 1.22.19 - ~\AppData\Local\Volta\tools\image\yarn\1.22.19\bin\yarn.CMD
    npm: 9.6.5 - ~\AppData\Local\Volta\tools\image\npm\9.6.5\bin\npm.CMD
    pnpm: 8.6.3 - ~\AppData\Local\Volta\tools\image\pnpm\8.6.3\bin\pnpm.CMD

### Describe the bug

I am not sure if it should end here or in next-auth repository, so I will link issues together.

### Link to reproduction

https://github.com/nextauthjs/next-auth/issues/8333

### To reproduce

Follow the steps from another issue.

pegak:
I found it!

All I had to do was copy the DB structure from Drizzle Docs - https://authjs.dev/reference/adapter/drizzle#postgres

There was a difference between T3 Turbo and Drizzle Docs with the naming of the tables (here we have `s` appended).

calebadcock:
Is there a better way for these to be defined? Seems odd to have these duplicated like they are now...

pegak:
To be honest I was confused (as SQL is not my strong skill), so I just copied from the docs.

currenthandle:
@BleedingDev Can you give some more specificity as to how you solved this? 

I copied `https://authjs.dev/reference/adapter/drizzle#mysql` I'm using Planetscale / MySQL to `schema/auth.ts` and reran `pnpm db:push`. It changed the table names to account, session, t3turbo_post, user, & verificationToken. 

But I am getting the same error as I was before (the one you described) `'ezkl-hub.t3turbo_account' doesn't exist`.

I can't find where this insertion is happening in the code so I've been pulling my hair out over this for a few hours.

Thanks in advance! 


I'm on commit `fd6f03bc96e0a9981be752f064930425b1466195`

currenthandle:
Ok I think I figured it out

Git diff:
```sh
-export const users = mySqlTable("users", {
[...]

PR #539, opened 2023-08-22
fix: remove appended `s` from table definitions

currenthandle:
This pull request resolves the issue with Drizzle Adapter record insertions described in this issue: https://github.com/t3-oss/create-t3-turbo/issues/511


juliusmarminge:
Sorry for not responding, I've been away on vacation and haven't touched Github for a week.

I'm catching up on some work and will get to this PR fixing this asap (i also need to migrate the db on the deployed app so that doesn't break - or maybe it is already broken :( )

PR #540, opened 2023-08-22
Update `expo-router` to use Expo 49 entry.

tomheaton:
Update `expo-router` to use Expo 49 entry.


tomheaton:
Investigating this breaking Android builds...

juliusmarminge:
> Investigating this breaking Android builds...

Yes, I think that was why i kept it the way it is. It worked fine locally but failed to build. Might be monorepo resolutions making the `expo-router/entry` not found when it's in package.json rather than imported?

juliusmarminge (approved):

PR #541, opened 2023-08-22
fix: turbo gen command

shoodey:
First of all, big love for all the work you have done!

## Issue

As it stands, the `pnpm turbo gen init` command throws a `SyntaxError: Cannot use import statement outside a module` error, disallowing the generation of packages.

## Solution

The changes aim to solve the issue by introducing `config.ts` and further enhancing the experience to include installing dev dependencies.

I tried to make the changes as atomic as possible so we can focus on fixing the error rather than introducing an extra feature.

PS: This was my first contribution to OSS, please let me know if you have any comments :)


turbo/generators/config.ts
```diff
@@ -51,30 +58,41 @@ export default function generator(plop) {
         path: "packages/{{ name }}/package.json",
         async transform(content, answers) {
           const pkg = JSON.parse(content);
+          // Add dependencies
```
claudiorivera, line 61: Since `answers` is `any` in this scope, we might wanna validate `answers`, like we do elsewhere:
```ts
if ("deps" in answers && typeof answers.deps === "string") {
  //...
}
```

turbo/generators/config.ts
```diff
@@ -16,6 +17,12 @@ export default function generator(plop) {
         message:
           "Enter a space separated list of dependencies you would like to install",
       },
+      {
+        type: "input",
+        name: "devdeps",
```
claudiorivera, line 22: I feel like this `devdeps` prompt might end up confusing people. For example, what if someone assumed that the `deps` prompt was meant to include all deps, including devDeps? Or maybe someone doesn't know (or care) about the difference between deps and devDeps? 🤷🏼 

I would ditch it, personally.

However, if we do want two prompts for deps, I would include something explicit in the first prompt like: `Enter a space separated list of dependencies you would like to install (you'll enter development dependencies next)`

claudiorivera (commented):
I almost submitted an identical PR 😅  Nice work!

turbo/generators/config.ts
juliusmarminge: Yea delete this prompt.

I had it initially but it's confusing since we already install some devDeps by default

shoodey: I see you point and removed change accordingly!

turbo/generators/config.ts
shoodey: Good catch, amended!

shoodey:
> I almost submitted an identical PR 😅 Nice work!

I'm just glad I beat you to it and having the opportunity to finally contribute :D

juliusmarminge (approved):

PR #547, opened 2023-08-24
Revert prettier & ts Configs to Global Namespace, Other/Minor Changes

peterkibuchi:
Breaking out `prettier` and `typescript` configs into their own packages is cool, but adds a decent amount of complexity to an already somewhat complex monorepo:
- It forces us to install prettier in 9 separate packages and maintain scripts for carrying out formatting with every new file type we add. We also need for a `.prettierignore` in each package for the files we want to ignore.
- Not to mention the files in the global namespace aren't formatted in the current setup, and to do so we would have to jump through extra hoops i.e. installing prettier in the global namespace (go figure), extending the base prettier config, and maintaining a separate script.
- Top-level `prettier` solves this as we can easily add/remove file types we want to format in one place, and format all the files in one go. Also, files in `.gitignore` are automatically ignored, so there's even no need for a `.prettierignore`.
- It just doesn't provide much benefit to have the packages do their own formatting, or at least not enough to justify all the complexity it introduces.

- Ditto for the base `typescript` config. Having it at the top level makes everything simpler. We actually _need_ to have it at the top level so we can include `prettier.config.mjs` without having to go through the extra steps of installing the `typescript` package in the top-level `package.json`, and extending it in yet another `tsconfig.json`.

Changes in this PR:
- Drop `prettier` and `typescript` packages, in favour of top level `prettier.config.mjs` & `tsconfig.json`, to reduce complexity; make corresponding changes across repo.
- Add separate CI step to `build-lint` to check if codebase is formatted correctly.
- Arrange VSCode recommended extensions so they appear in alphabetical order in the code editor.
- Format and make minor tweaks across repo to improve readability and overall aesthetic of the codebase.


juliusmarminge:
Sorry but having them separately was made for a reason, utilise parallelism with turbo as much as possible. the ci is quite a bit faster with this setup.

making the root formatted can be easily fixed while retaining the tooling packages

PR #711, opened 2023-10-30
Update Solito wording

nandorojo:
Glad Solito is on the FAQ here. I changed the wording slightly – the existing wording made it sound like *you can't use Solito with T3*. Reading the paragraph would clarify that missed assumption, but I thought it would be best to update the heading.

Nice work.


juliusmarminge (approved):
Thank you!

PR #735, opened 2023-11-08
Update base tsconfig.json

mattpocock:
Updated to match recommendations in: https://www.totaltypescript.com/tsconfig-cheat-sheet


juliusmarminge (approved):
Thanks Matt!

PR #749, opened 2023-11-11
Add prettierignore

dBianchii:
If you run `turbo build`, all subsequent `turbo format` commands fail.

This PR fixes this problem.

Before prettierignore:
image

After prettierignore:
image


juliusmarminge (approved):
Could we do `--ignore-path ../../.gitignore` instead? Kinda hate these small files everywhere 😅

dBianchii:
I was trying to do this, with `--ignore-path='**/.next/**' --ignore-path='**/.nitro/**`, but had no luck. I didn't know you could specify a .gitignore file like this. The solution is much better

Btw julius, I sent you a small sponsorship on github, to support all of the amazing stuff you've been doing. Keep up the amazing work!

dBianchii:
Do you think we should keep these package.json files consistent?
Most of them have no build output. Might be better to help keep consistency
image

juliusmarminge:
> Btw julius, I sent you a small sponsorship on github, to support all of the amazing stuff you've been doing. Keep up the amazing work!

I noticed, thank you 🙏

juliusmarminge:
> Do you think we should keep these package.json files consistent?
> 
> Most of them have no build output. Might be better to help keep consistency
> 
> image
> 

Sure, might be able to remove the extension matcher and just format everything that isn't gitignored?

prettier . --ignore-path ...

juliusmarminge:
I should also do `type: module` so we dont need mjs files and just use .cjs for the few exceptions (like postcss.config which nextjs cant handle as esm)

apps/auth-proxy/package.json
```diff
@@ -7,7 +7,7 @@
     "clean": "rm -rf .turbo node_modules",
     "dev": "nitro dev --port 3001",
     "lint": "eslint .",
-    "format": "prettier --check \"**/*.{mjs,ts,md,json}\"",
+    "format": "prettier --check \"**/*.{mjs,ts,md,json}\" --ignore-path ../../.gitignore",
```
juliusmarminge, line 10: rm match pattern

PR #794, opened 2023-12-14
fix: prettier overwrites @typescript-eslint/consistent-type-imports  

noxify:
There is/was a misconfiguration (?) between prettier and eslint.

The eslint rule `@typescript-eslint/consistent-type-imports` has updated something like this ( while running `pnpm lint:fix`):

```ts
import Markdoc, { type Node as MarkdocNode, type Tag } from "@markdoc/markdoc"
```

to:

```ts
import type { Node as MarkdocNode, Tag } from "@markdoc/markdoc"
import Markdoc from "@markdoc/markdoc"
``` 

After running `pnpm format:fix` ( or just save a linted file ), prettier has set the import to the old value.

After setting `<TYPES>` and `<TYPES>^[.]"` in the prettier config the `pnpm format:fix` returns the correct import ( = same as `pnpm lint:fix` ).

Doc ref: https://github.com/IanVS/prettier-plugin-sort-imports?tab=readme-ov-file#4-group-type-imports-separately-from-values


juliusmarminge (changes requested):
This changes the import order and puts our own modules at the top amonst third party modules. I prefer 3rd party stuff to be at the top, then @acme/* then "nromal" imports:

```ts
// 1. 3rd party
import type { X } from "x";
import { x } from "x";
import { y } from "y";

// 2. then local packages
import type { AppRouter } from "@acme/api";
import { foo } from "@acme/foo";

// 3. then imports within the package
import type { Bar } from "~/foobar";
import {} from "./whatever";
...
```

noxify:
Thanks @juliusmarminge for the quick review.

I have updated the import order to support the described behavior.

juliusmarminge:
Awesome, can you also run the formatter?

packages/auth/index.ts
```diff
@@ -1,8 +1,7 @@
-/* eslint-disable @typescript-eslint/unbound-method */
```
noxify, line 1: fixes the eslint warning

```
Unused eslint-disable directive (no problems were reported from '@typescript-eslint/unbound-method').
```

noxify:
> Awesome, can you also run the formatter?

done - fixed also the found linting issue/warning

juliusmarminge (approved):

PR #795, opened 2023-12-14
chore(expo) : Update to Expo 50

### LINKED ISSUE #849: feat: Disallow pnpm workspaces hoisted node_modules packages to be used by all other 'child' sub @acme packages
### Describe the feature you'd like to request

I came across this problem multiple times, where... if you have a dependency inside `@acme/packageA` and then you have a `@acme/packageB` which depends on `@acme/packageA`, then immediatelly all `@acme/packageA` dependencies are allowed to be used and imported by `@acme/packageB`. This makes the project slightly harder to understand, and also breaks some dependecy principles.

I believe this is being caused by pnpm workspaces, where it will hoist all node_modules to the top, and somehow allow for this to happen.

In my opinion, if I install 'dayjs' inside `@acme/api` for example, then it shouldn't be accsessible in `@acme/nextjs` until I explicitly install it for this package

### Describe the solution you'd like to see

Some kind of configuration to disallow this behaviour

### Additional information

You can see an example of this in my forked project, where I have momentjs being imported in a nextjs application, that has no momentjs explicitly depended upon.
https://github.com/dBianchii/kodix-turbo/blob/main/apps/kdx/src/app/_components/apps/calendar/create-event-dialog.tsx#L6
https://github.com/dBianchii/kodix-turbo/blob/main/apps/kdx/package.json#L16

### LINKED ISSUE #850: feat: update to Expo SDK 50
### Describe the feature you'd like to request

Expo 50 has among other niceties better support for pnpm and turborepo

### Describe the solution you'd like to see

- update to sdk 50
- update expo router to v3
- update react-native?
- remove .npmrc

### Additional information

Closes #849

VaniaPopovic:
Closes #850
Closes #849

- Update dependancies for expo app to v5.
- Remove node-linker=hoisted as expo 50 plays nicer with pnpm
-  Change metro config based on @byCedric 's work in https://github.com/byCedric/expo-monorepo-example/pull/107
- Update expo router to v3 and change the required files 

Currently there's an issue with NativeWind failing to import react-native-css-interop/jsx-runtime that it depends on. Will raise an issue in Nativewind's repo and update this PR with the status. 
image

Nativewind issue to track : https://github.com/marklawlor/nativewind/issues/701


.npmrc
juliusmarminge, line 1: can we nuke this file all together now? I think strict-peer-dependencies was also set for expo reasons

apps/expo/expo-plugins/with-modify-gradle.js
juliusmarminge, line 1: 👀 not needed anymore?

apps/expo/index.js
juliusmarminge, line 1: this shouldn't be needed afaik?

apps/expo/package.json
```diff
@@ -2,10 +2,10 @@
   "name": "@acme/expo",
   "version": "0.1.0",
   "private": true,
-  "main": "expo-router/entry",
```
juliusmarminge, line 5: this should work according to their docs?

apps/expo/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "@acme/tsconfig/base.json",
+  "extends": ["@acme/tsconfig/base.json", "expo/tsconfig.base"],
```
juliusmarminge, line 2: change order here?

apps/expo/package.json
byCedric: This should work fine. We did discover an issue in `expo/AppEntry` where we try to import `../../App`, which doesn't work for `.pnpm/expo@.../node_modules/expo/AppEntry.js`.

We know about that issue, and will fix it as soon as we can!

.npmrc
byCedric: You can on the native side! We've been working hard to get rid of all implicit dependency references in both `/android` and `/ios` folders.

That means things like `expo run android|ios` are working as intended, without `node-linker=hoisted`.

Unfortunately, there are a few left-over issues, e.g. `eas update` won't work properly with this yet. Working hard to resolve them before SDK 50 stable is released.

apps/expo/index.js
byCedric: Exactly, you can, but you don't need to. The only reason why I did this is to let Turborepo determine if the Metro cache can be reused.

This was mostly due to the fact that there was no environment variable support, and therefore Metro did reuse the cache incorrectly when this was changed.

But, now that we have `@expo/env` and proper environment variables support, this shouldn't be necessary anymore.

juliusmarminge:
Are there still blockers now thst 50 is properly released?

dBianchii:
Hi, does this intend to be reopened? @VaniaPopovic ? Or maybe we can create a new PR

.npmrc
juliusmarminge, line 1: can we remove this entirely? idk when the strict-peer-dependency setting was introduced but ideally we wouldn't need it anymore

apps/expo/index.js
juliusmarminge, line 1: the example on expo uses the package.json field, should be fine to keep that right? https://github.com/expo/expo/blob/main/templates/expo-template-tabs/package.json#L3

idk what the main reason was for android not to work with that but if their pnpm support has gotten better i assume this issue might be fixed?

juliusmarminge: saw this comment, idk how much progress they'vem ade since? https://github.com/t3-oss/create-t3-turbo/pull/795/files#r1432898627

apps/expo/src/app/_layout.tsx
```diff
@@ -1,13 +1,18 @@
...
 
 import "../styles.css";
 
+import { useColorScheme } from "nativewind";
+
 // This is the main layout of the app
 // It wraps your pages with the providers they need
 export default function RootLayout() {
+  const { colorScheme } = useColorScheme();
+  const insets = useSafeAreaInsets();
```
juliusmarminge, line 15: Unused?

apps/expo/src/app/index.tsx
```diff
@@ -97,31 +96,37 @@ function CreatePost() {
...
   const utils = api.useUtils();
 
   const postQuery = api.post.all.useQuery();
 
   const deletePostMutation = api.post.delete.useMutation({
-    onSettled: () => utils.post.all.invalidate(),
+    onSettled: () => utils.post.all.invalidate().then(),
   });
 
+  useEffect(() => {
```
juliusmarminge, line 108: Can't this be in a root / set in some provider instead of an effect?

apps/expo/tailwind.config.ts
```diff
@@ -2,9 +2,50 @@ import type { Config } from "tailwindcss";
 // @ts-expect-error - no types
 import nativewind from "nativewind/preset";
 
-import baseConfig from "@acme/tailwind-config";
```
juliusmarminge, line 5: Why?

VaniaPopovic: Nativewind does not support some of the features we are using in the nextjs app, namely  ```  borderRadius: {
        lg: "var(--radius)",
        md: "calc(var(--radius) - 2px)",
        sm: "calc(var(--radius) - 4px)",
      }, ```

You can see more https://github.com/marklawlor/nativewind/issues/764 here

apps/expo/src/app/_layout.tsx
VaniaPopovic: colorScheme is used there, to pass the colours to expo-status-bar which uses RGB colours that we define in ``styles.css```.   I used RGB colours because we cannot read the css variables there https://github.com/marklawlor/nativewind/issues/682

apps/expo/tailwind.config.ts
juliusmarminge: Maybe split the preset up then to core/web/native?

apps/expo/src/app/index.tsx
VaniaPopovic: Removed it entirely as it is not needed

juliusmarminge (approved):

PR #799, opened 2023-12-17
Use planetscale `connect` instead of `Client` to allow additional configuration

Itsnotaka:
use planetscale connect instead of Client to allow additional configuration

### DIFF
--- a/packages/db/index.ts
+++ b/packages/db/index.ts
@@ -1,4 +1,4 @@
-import { Client } from "@planetscale/database";
+import { connect } from "@planetscale/database";
 import { drizzle } from "drizzle-orm/planetscale-serverless";
 
 import * as auth from "./schema/auth";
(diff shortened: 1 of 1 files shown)


packages/db/index.ts
```diff
@@ -10,11 +10,10 @@ export { mySqlTable as tableCreator } from "./schema/_table";
...
-export const db = drizzle(
-  new Client({
-    host: process.env.DB_HOST,
-    username: process.env.DB_USERNAME,
-    password: process.env.DB_PASSWORD,
-  }).connection(),
-  { schema },
-);
+const connection = connect({
+  host: process.env.DB_HOST!,
```
juliusmarminge, line 14: use `env` from `~/env`

juliusmarminge (changes requested):

packages/db/index.ts
juliusmarminge: oh wait my bad - thought this was on create-t3-app

juliusmarminge (approved):

PR #812, opened 2023-12-24
Polish PR: Minor Refactoring & Formatting Improvements

peterkibuchi:
This is a "polish up" PR. I've split it into two commits to ease code review, and to ensure if changes are requested it won't block the entire PR.

Changes in 1st Commit:
- Organize variables and scripts across packages alphabetically
- Replace broken link in `.github/ISSUE_TEMPLATE/feature_request.yml`
- Edit the `typecheck` job in `.github/workflows/ci.yml` to have pnpm running it instead of turbo
- Minor formatting improvements

Changes in 2nd Commit:
- Add `prettier` to the `eslint` package; it currently doesn't have it installed
- Remove unnecessary `React` import from `apps/expo/src/app/_layout.tsx`
- Standardize function format across `expo` app
- Import `useState` instead of the whole `React` object
- Organize recommended extensions so they appear in the Extensions Tab in alphabetical order

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,9 +1,9 @@
 {
   "recommendations": [
-    "bradlc.vscode-tailwindcss",
     "dbaeumer.vscode-eslint",
-    "esbenp.prettier-vscode",
     "expo.vscode-expo-tools",
-    "yoavbls.pretty-ts-errors"
+    "esbenp.prettier-vscode",
+    "yoavbls.pretty-ts-errors",
+    "bradlc.vscode-tailwindcss"
   ]
 }
--- a/apps/expo/package.json
+++ b/apps/expo/package.json
@@ -8,11 +8,11 @@
     "dev": "expo start --ios",
     "dev:android": "expo start --android",
     "dev:ios": "expo start --ios",
-    "lint": "eslint .",
-    "format": "prettier --check . --ignore-path ../../.gitignore",
-    "typecheck": "tsc --noEmit",
     "android": "expo run:android",
-    "ios": "expo run:ios"
+    "ios": "expo run:ios",
+    "format": "prettier --check . --ignore-path ../../.gitignore",
+    "lint": "eslint .",
+    "typecheck": "tsc --noEmit"
   },
   "dependencies": {
     "@expo/metro-config": "^0.10.7",
--- a/apps/expo/src/app/index.tsx
+++ b/apps/expo/src/app/index.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import { useState } from "react";
 import { Button, Pressable, Text, TextInput, View } from "react-native";
 import { SafeAreaView } from "react-native-safe-area-context";
 import { Link, Stack } from "expo-router";
@@ -39,8 +39,8 @@ function PostCard(props: {
 function CreatePost() {
   const utils = api.useUtils();
 
-  const [title, setTitle] = React.useState("");
-  const [content, setContent] = React.useState("");
+  const [title, setTitle] = useState("");
+  const [content, setContent] = useState("");
 
   const { mutate, error } = api.post.create.useMutation({
     async onSuccess() {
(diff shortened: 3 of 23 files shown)


.vscode/extensions.json
```diff
@@ -1,9 +1,9 @@
 {
   "recommendations": [
```
juliusmarminge, line 2: So this is sorted based on the name after the . ?

apps/expo/package.json
```diff
@@ -4,15 +4,15 @@
   "private": true,
   "main": "expo-router/entry",
   "scripts": {
+    "android": "expo run:android",
+    "ios": "expo run:ios",
```
juliusmarminge, line 8: This seems out of place

tooling/eslint/nextjs.js
```diff
@@ -3,7 +3,6 @@ const config = {
   extends: ["plugin:@next/next/recommended"],
   rules: {
     "@next/next/no-html-link-for-pages": "off",
-    "@typescript-eslint/require-await": "off",
```
juliusmarminge, line 6: This rule is deactivated as server actions requires to be async but they might not have await statements

apps/expo/src/utils/api.tsx
```diff
@@ -43,10 +42,9 @@ const getBaseUrl = () => {
  * A wrapper for your app that provides the TRPC context.
  * Use only in _app.tsx
  */
-
-export function TRPCProvider(props: { children: React.ReactNode }) {
-  const [queryClient] = React.useState(() => new QueryClient());
-  const [trpcClient] = React.useState(() =>
+export function TRPCProvider({ children }: { children: React.ReactNode }) {
```
juliusmarminge, line 45: Opinionated, but for one-odds i dont like destructuring props like this. Revert

apps/expo/src/app/index.tsx
```diff
@@ -7,29 +7,31 @@ import { FlashList } from "@shopify/flash-list";
...
 import { api } from "~/utils/api";
 
-function PostCard(props: {
+interface PostCardProps {
   post: RouterOutputs["post"]["all"][number];
   onDelete: () => void;
-}) {
+}
+
+function PostCard({ post, onDelete }: PostCardProps) {
```
juliusmarminge, line 15: Same, revert destructuring.

Also dont see why the type has been brought out to an interface rather than inlined, as it is inlined everywhwre Else

juliusmarminge (commented):
Mostly good

.vscode/extensions.json
peterkibuchi: Not quite. It's based on the name of the extension as seen in the (recommended) extensions tab. It makes sense to arrange them alphabetically there because that's what most users will interact with, as opposed to alphabetically within this file.

juliusmarminge: I really dont care, just thinking of future maintenance. What order will new addtions be placed in?

peterkibuchi: The order in which they appear in the extensions tab?

juliusmarminge: eehm am i missing something here or is the extensions tab sorted according to the array? 

https://github.com/t3-oss/create-t3-turbo/assets/51714798/a68d63c1-aa8f-4a0f-a87e-fe70268cbece

peterkibuchi: You're correct — the extensions appear in the tab in the order in which they appear in the array. That's why I thought it'd be neat to sort the array so they appear in alphabetical order in the tab. See:

Screenshot from 2023-12-28 05-33-48

I know it's a minor, inconsequential change, but I thought it'd be a nice touch. Recommended extensions aren't updated regularly anyway, so "maintaining" it won't be a problem.

But if you find it unnecessary, I can revert the changes.

juliusmarminge: So my first comment stands then 😅

> So this is sorted based on the name after the . ?

juliusmarminge (approved):

PR #814, opened 2023-12-29
chore: update trpc to use server caller

juliusmarminge:
moves server caller to use `createCaller` now that we've added the ability to pass it a createContext function

> [!NOTE]  
> I will wait for PR (https://github.com/trpc/trpc/pull/5244) to be merged then revert to using the `next` tag


juliusmarminge:
cc @RhysSullivan iirc you preffered the server caller approach so I think you'll like this change. Reason I didn't like this before was the nested awaits required when you had async context (see https://github.com/trpc/trpc/pull/5244#issue-2058751912) , but that's fixed by allowing the createCaller function be passed a function

PR #832, opened 2024-01-05
feat: use expo-linking to construct urls for expo auth

### LINKED ISSUE #486: Expo authentication using next-auth

jmcmullen:
### Summary
This PR adds compatibility for Next-Auth/Authjs with the Expo app.

Thanks to @juliusmarminge for doing most of the work in #720.
Closes #486 

In order to get it working locally, you will need to set the AUTH_URL environment variable to your local IP address instead of localhost otherwise CRSF errors will be encountered. This isn't a massive concern since you will need to edit the baseUrl logic before deploying to production anyway.

### Demo:
https://github.com/t3-oss/create-t3-turbo/assets/6767542/285ab2ae-4efb-47fc-b5ca-c590a77ceda9


apps/nextjs/src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -2,47 +2,47 @@ import type { NextRequest } from "next/server";
...
     });
   }
 
   if (nextauthAction === "callback" && !!isExpoCallback) {
-    cookies().delete(expoRedirectCookieName);
+    cookies().delete(EXPO_COOKIE_NAME);
 
-    const authResponse = await _GET(req);
+    const authResponse = await DEFAULT_GET(req);
     const setCookie = authResponse.headers.getSetCookie()[0];
```
Comment (mateuszdziuba) [inline apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:35]: I was trying to use this solution with Google and it didn't work, after some debugging it turned out this cookie is actually at index 1, so I think this part needs to be a find or something

Issue #867, opened 2024-01-29
bug: EAS build not working with hoisting


EuanMorgan:
### Provide environment information

  System:
    OS: macOS 14.2.1
    CPU: (8) arm64 Apple M2
    Memory: 179.27 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - /usr/local/bin/node
    Yarn: 1.22.21 - ~/Library/pnpm/yarn
    npm: 10.2.4 - /usr/local/bin/npm
    pnpm: 8.15.0 - /usr/local/bin/pnpm
    Watchman: 2023.12.04.00 - /opt/homebrew/bin/watchman

### Describe the bug

After the recent Expo 50 upgrade, running EAS build no longer works.

This is the error message:
`Cannot find 'expo-modules-autolinking' package in your project, make sure that you have 'expo' package installed`

I think it's because of the removal of .npmrc and the hoisting changes,
if I re-add `node-linker=hoisted` it works.

### Link to reproduction

https://github.com/EuanMorgan/expo-50-test

### To reproduce

Try to run an EAS build:
`pnpm i`
`cd apps/expo`
`eas build -p ios`

juliusmarminge:
Hoisting should work in Expo 50.

cc @VaniaPopovic mind having a look?

akashdevcc:
Why this issue only happens with `eas build -p ios` and not with `eas build -p android`?

PR #868, opened 2024-01-30
fix(expo): Revert the removal of .npmrc

VaniaPopovic:
This PR reverts the removal of the .npmrc file because eas-cli doesn't play well with pnpm yet. 
See more at https://github.com/t3-oss/create-t3-turbo/issues/867


.npmrc
```diff
@@ -0,0 +1,2 @@
+# See https://github.com/t3-oss/create-t3-turbo/issues/867
+node-linker=hoisted
```
juliusmarminge, line 2: Can we get away with public-hoist-pattern?

juliusmarminge:
make it a single commit please

PR #877, opened 2024-02-05
feat: hydration helpers [proof of concept]

### LINKED ISSUE #876: feat: Hydrate Client Component (QueryClient) state w/ dehydrate in RSC
### Describe the feature you'd like to request

Instead of running tRPC queries directly in the RSC and passing it's result to the client component as props, we should prefetch the data and hydrate QueryClient data to the client components instead.

Right now data is fetched in the RSC like so:
https://github.com/t3-oss/create-t3-turbo/blob/7c1da52c6fdfe108d8b9644047308c3e75a58214/apps/nextjs/src/app/page.tsx#L15

This data is passed to the client component via props, however the query is then run again here
https://github.com/t3-oss/create-t3-turbo/blob/7c1da52c6fdfe108d8b9644047308c3e75a58214/apps/nextjs/src/app/_components/posts.tsx#L89-L91

**Note**: Using `initialData` isn't actually what we want. This adds data to the cache, only if something doesn't already exist. So if you added another page and you navigated between them... if your data changes (in the background), the data loaded via the RSC won't change (as it's old data is still cached via `initialData`).

Right now data is fetched in the RSC (passed to the client via props), and then again on the client (double query.. which isn't what we want).

Instead, we should hydrate the QueryClient via the RSC. This guarantees that fresh data will always be available to the client component.

### Describe the solution you'd like to see

I've got something like this working in my project
```ts
import { Suspense } from "react";
import { appRouter } from "@acme/api";
import { dehydrate, HydrationBoundary } from "@tanstack/react-query";
import { createServerSideHelpers } from "@trpc/react-query/server";
import superjson from "superjson";

import { createContext } from "~/trpc/server";
import { CreatePostForm, PostList } from "./components/posts";
import Spinner from "./components/spinner";

async function Posts() {
  const api = createServerSideHelpers({
    router: appRouter,
    ctx: await createContext(),
    transformer: superjson,
  });
  await api.post.all.prefetch();
  const dehydratedState = dehydrate(api.queryClient);

  return (
    <HydrationBoundary state={dehydratedState}>
      <PostList />
    </HydrationBoundary>
  );
}

export default async function HomePage() {
  return (
    <>
      <CreatePostForm />
      <Suspense fallback={<Spinner />}>
        <div className="w-full max-w-2xl overflow-y-scroll">
          <Posts />
        </div>
      </Suspense>
    </>
  );
}
```

The `PostList` component can then be rendered like so
```ts
export function PostList() {
  const [posts] = api.post.all.useSuspenseQuery();

  if (posts.length === 0) {
    return (
      <div className="relative flex w-full flex-col gap-4">
        <PostCardSkeleton pulse={false} />
        <PostCardSkeleton pulse={false} />
        <PostCardSkeleton pulse={false} />
        <div className="absolute inset-0 flex flex-col items-center justify-center bg-black/10">
          <p className="text-2xl font-bold text-white">No posts yet</p>
        </div>
      </div>
    );
  }

  return (
    <div className="flex w-full flex-col gap-4">
      {posts.map((p) => {
        return <PostCard key={p.id} post={p} />;
      })}
    </div>
  );
}
```

Also it should be noted that `refetchOnMount` is set to `false` so queries are never run on mount. Remember.. live data is being hydrated in the RSC.. so there's no reason to re-run the queries on the client side.

```ts
export const queryClientOptions = {
  defaultOptions: {
    queries: {
      refetchOnMount: false,
    },
  },
};

// trpc/client.tsx
const [queryClient] = useState(() => new QueryClient(queryClientOptions));
```

👍🏻 

### Additional information

_No response_

juliusmarminge:
Closes #876


apps/nextjs/src/app/page.tsx
```diff
@@ -33,10 +32,21 @@ export default async function HomePage() {
...
           </Suspense>
         </div>
       </div>
     </main>
   );
 }
+
+async function PostListWrapped() {
+  const posts = await api.post.all();
+  setQueryData((t) => [t.post.all, undefined], posts); // would be cool if this was automatically put to the cache
```
juliusmarminge, line 45: this could be abstracted away by a `createCaller` from `@trpc/react-query` that would automatically set data to a query client cache, then the user would just need to put in `<HydrateClient>` blocks whereever they need to have live, updated data in client components

apps/nextjs/src/app/page.tsx
```diff
@@ -33,10 +32,21 @@ export default async function HomePage() {
...
+            <PostListWrapped />
           </Suspense>
         </div>
       </div>
     </main>
   );
 }
+
+async function PostListWrapped() {
+  const posts = await api.post.all();
```
KATT, line 44: 1. Maybe the result of `api.post.all()` could be _branded_ somehow so we know the or can figure out the full query key

apps/nextjs/src/app/page.tsx
```diff
@@ -33,10 +32,21 @@ export default async function HomePage() {
...
+  setQueryData((t) => [t.post.all, undefined], posts); // would be cool if this was automatically put to the cache
+
+  return (
+    <HydrateClient>
```
KATT, line 48: 2. If it was `posts` were branded, we could do some helper that automatically knew how to add it in the RQ cache, like

```tsx
<HydrdateClient queries={[ posts ]}>
   <PostList />
</HydrateClient>
```

KATT: Or maybe it could be something altogether automagic similar to https://tanstack.com/query/latest/docs/framework/react/guides/advanced-ssr#experimental-streaming-without-prefetching-in-nextjs

and we'd just have a root `<Hydrate>` thing in the layout that somehow had a stream of data it pushed to the cache (not sure if it's technically feasible unless we can hack RSC<->client comms to work with streams or async iterators

PR #919, opened 2024-03-12
feat: cache `auth` in RSC

dBianchii:
This PR changes the `auth()` export in @acme/api to go through react/cache. This will allow for de-duplicating of session calls and will only do one per request across all components


dBianchii:
I wonder if we should do an example of calling auth() in 2 separate RSCs, to showcase how you don't need to pass session around every time

dBianchii:
@juliusmarminge after we approve this, let me know if I can do the same over on create-t3-app. I think it's also not using react/cache with the app router

packages/auth/src/index.ts
```diff
@@ -37,3 +38,10 @@ export const {
     },
   },
 });
+
+export { GET, POST, signIn, signOut };
+/**
+ * This is the main way to get session data for your RSCs.
+ * This will de-duplicate all calls to next-auth's default `auth()` function and only call it once per request across all components
+ */
+export const auth = cache(defaultAuth);
```
juliusmarminge, line 47: This doesn't work in middleware does it?

dBianchii: In Next's middleware.ts?

dBianchii: Then I believe we could add a comment somewhere, or in Readme.md explaining to also export `defaultAuth`  for that use case. What do you suggest?

juliusmarminge: i think we should have a middleware.ts file by default tbh

dBianchii: ¯\_(ツ)_/¯ in this case yea sure

Just thought that this repo was supposed to be a "turbo" version of ct3a. And since in there we don't specify middleware files, don't see why we should also do it here? Could be a good idea though.

juliusmarminge: afaik middleware is the only way to make the session refresh?

juliusmarminge (approved):

dBianchii:
Wait, but with this setup how would you import from index.rsc? I see that `auth-showcase.tsx` is still efectively importing from index.tsx only

packages/auth/package.json
```diff
@@ -4,7 +4,10 @@
   "private": true,
   "type": "module",
   "exports": {
-    ".": "./src/index.ts",
+    ".": {
+      "react-server": "./src/index.rsc.ts",
```
juliusmarminge, line 8: This here ^^

dBianchii:
> Exports condition

Cntrl + click on auth-showcase.ts auth() still is bringing me to index.ts tho, not index.rsc.ts

juliusmarminge:
yea the default entrypoint is still the normal index file.

exports condition are evaluated at runtime based on the thing that imports it

juliusmarminge:
See https://github.com/reactjs/rfcs/blob/main/text/0227-server-module-conventions.md#react-server-conditional-exports

Issue #636, opened 2023-09-22
enhancement: extend `localhost`-getter beyond dev lifecycle


angelprzz:
### Provide environment information

  System:
    OS: macOS 13.5.2
    CPU: (10) arm64 Apple M1 Max
    Memory: 341.61 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.17.1 - ~/.nvm/versions/node/v18.17.1/bin/node
    npm: 9.6.7 - ~/.nvm/versions/node/v18.17.1/bin/npm

### Describe the bug

This is a similar problem @johnslemmer  had a few months ago (https://github.com/t3-oss/create-t3-turbo/pull/300).

Basically, if you build and run your app using `expo run:ios`, the hostUri becomes unaccessible. Manifest & Manifest2 are also null (and deprecated). 

There must be a way to get the localhost address without Expo Go.

### Link to reproduction

https://github.com/t3-oss/create-t3-turbo/pull/300

### To reproduce

1. Create a t3 turbo project and set db & server.
2. Inside the expo folder, run `npx expo run:ios` and wait for the project to build inside the simulator
3. At this moment, `Constants.expoConfig.hostUri` is null and there's no way to access the local host.

### Additional information
Expo 49.0.11

_No response_

johnslemmer:
FYI

I think in expo 49 some more things changed. That line to figure out `debuggerHost` now looks like this in my project.

```js
  const debuggerHost =
    Constants.expoConfig?.hostUri ??
    Constants.manifest2?.extra?.expoGo?.debuggerHost;
```

Hope this helps.

angelprzz:
Thank you @johnslemmer 
Unfortunately the debuggerHost is still undefined when running the app on bare workflow.
I'm using Expo 49.0.11, but I had the same problem on previous versions too.

juliusmarminge:
Open for improvements here - I'm not really in the Expo ecosystem a whole ton so unsure what best practices / all the APIs are...

The current solution was meant to work for dev and then you'd point it to an actual URL when building - but if there is a way to extend the getter to work for other lifecycle scripts I'm open to include those!

angelprzz:
@juliusmarminge Yeah, I'm trying to find a solution too. 
The reason why this is important is because in order to test things like push notifications you need to run the app locally with `npx expo run:ios` on a physical device, so Expo Go wouldn't work in these kind of situations.

[...]

PR #938, opened 2024-03-25
fix(deps): Integrate expo-dev-client for enhanced capabilities in bare builds (#636)

angelprzz:
Check issue #636 

I realized that this problem can be solved by installing the `expo-dev-client` package. This package brings some of the capabilities of Expo Go to bare projects, including the ability to access the debuggerHost.

I tested it on both Android and iOS builds and works perfectly. This should be enough to solve the issue.

Additionally, I added the `/ios` and `/android` folders to `.gitignore`, so the are not committed after prebuilding the project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@ next-env.d.ts
 dist/
 expo-env.d.ts
 apps/expo/.gitignore
+apps/expo/ios
+apps/expo/android
 
 # production
 build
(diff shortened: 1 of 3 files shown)


angelprzz:
@juliusmarminge @johnslemmer  Please guys, take a look. It's an easy fix that fixes lots of problems.

juliusmarminge (approved):

PR #951, opened 2024-03-28
NextJS: Avoid immediately refetching on the client. 

hoangtrung99:
NextJs app:
 Add `staleTime` to the `queryClient` to avoid immediately refetching on the client.


apps/nextjs/src/trpc/react.tsx
```diff
@@ -8,7 +8,16 @@ import SuperJSON from "superjson";
...
 import type { AppRouter } from "@acme/api";
 
-const createQueryClient = () => new QueryClient();
+const createQueryClient = () =>
+  new QueryClient({
+    defaultOptions: {
+      queries: {
+        // With SSR, we usually want to set some default staleTime
+        // above 0 to avoid refetching immediately on the client
+        staleTime: 60 * 1000,
```
juliusmarminge, line 17: isn't 60 a bit much? The next.js router i think uses 30 sec

hoangtrung99: Thank you for your review. I think it will depend on each project, however, 30 seconds might be suitable as the default. I have updated it.

juliusmarminge: Yup ofc will differ on project to project. 30sec is what I had in mind for #877. I think it's a bit more "sane default"

juliusmarminge (approved):

PR #954, opened 2024-03-31
Move `api` to `@acme/api`

ahkhanjani:
By moving this line:

```ts
export const api = createTRPCReact<AppRouter>();
```

from `apps/nextjs/src/trpc/react.tsx` to `packages/api` we can make it usable in other React based packages that rely on API calls. Also reduces duplication in Next.js packages.


juliusmarminge:
Now were making the api depend on trpc/client and React query which i dont think makes much sense

If you want to call the api from some package maybe you should make some "React" package where you share react hooks etc and then it might make sense to have this in there

PR #956, opened 2024-04-03
feat: shadcn-ui cli working in packages/ui

trevorpfiz:
I was able to follow https://github.com/dan5py/turborepo-shadcn-ui and get the `shadcn-ui` cli working in `packages/ui`. 

The only catch is the folder structure of `packages/ui` is not as clean, but it follows the standard pattern. Also had to add `"@acme/ui/*": ["../../packages/ui/src/*"]` to the nextjs `tsconfig.json` as outlined in https://github.com/dan5py/turborepo-shadcn-ui/issues/20

Seems like the pros outweigh the cons though.


juliusmarminge:
Thanks for the PR. However I don't want to move all the components 2 levels deep for "no" reason. I'll just add the workaround with the self-referncing to `cn()` in the orginal PR

PR #969, opened 2024-04-07
chore: update to eslint 9

juliusmarminge:


packages/ui/eslint.config.js
```diff
@@ -0,0 +1,8 @@
+// FIXME: This kinda stinks...
+/// <reference types="../../tooling/eslint/types.d.ts" />
```
ixahmedxi, line 2: that does stink

juliusmarminge: fixed it

.vscode/settings.json
```diff
@@ -5,6 +5,7 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   "eslint.rules.customizations": [{ "rule": "*", "severity": "warn" }],
+  "eslint.experimental.useFlatConfig": true,
```
juliusmarminge, line 8: looks like the new version is kinda close https://github.com/microsoft/vscode-eslint/releases

RiccardoRomagnoli:
@juliusmarminge eslint-plugin-import@2.29.1 does not support eslint 9

PR #1012, opened 2024-05-03
fix: Update Expo SDK version in README.md

mrvndss:


juliusmarminge (approved):
should upgrade to sdk51 but this'll do for now :)

PR #1038, opened 2024-05-16
feat: multi-root workspaces

trevorpfiz:
https://code.visualstudio.com/docs/editor/multi-root-workspaces

I came across [next-fast-turbo](https://github.com/cording12/next-fast-turbo) when adding a FastAPI to a Turborepo, and I enjoyed the development experience of using multi-root workspaces. https://next-fast-turbo.mintlify.app/documentation/local-development#working-with-a-monorepo-in-vs-code

Any experience working with multi-root workspaces? One issue I noticed is that there is some overlap between the root `launch.json` configurations and the ones defined in `.code-workspace`. Leads to this when working in the multi-root workspace:

image

I don't have enough experience to say, but we could recommend multi-root workspaces in the README and then remove the root `launch.json`, but only if there are clear advantages to working with Turborepo in a multi-root workspace.

### DIFF
--- a/.vscode/create-t3-turbo.code-workspace
+++ b/.vscode/create-t3-turbo.code-workspace
@@ -1,0 +1,35 @@
+{
+  "folders": [
+    {
+      "name": "Next.js",
+      "path": "../apps/nextjs",
+    },
+    {
+      "name": "Expo",
+      "path": "../apps/expo",
+    },
+    {
+      "name": "Root",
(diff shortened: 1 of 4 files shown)


juliusmarminge:
I have no personal experience with this so doesn't feel comfortable merging this.

Maybe start a discussion and see what others think and maybe we can add it in the future if the response is positive

PR #1054, opened 2024-05-28
feat: Expo Auth without setting AUTH_URL.

Wundero:
This PR implements logic into the Expo <-> NextAuth linking which enables the IP that expo uses (e.g. 192.168.x.y) to work without having to set the AUTH_URL to be that IP.

See #720 for details and discussion related to this issue.


apps/nextjs/src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -1,18 +1,42 @@
...
 const EXPO_COOKIE_NAME = "__acme-expo-redirect-state";
 const AUTH_COOKIE_PATTERN = /authjs\.session-token=([^;]+)/;
 
+/**
+ * Correct request.url for local development so that Expo can work. Does nothing in production.
+ * @param req The request to modify
+ * @returns The modified request.
+ */
+function rewriteRequestUrl(req: NextRequest) {
+  if (env.VERCEL) {
```
Comment (ochicf) [inline apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:19]: Should the case of an app not being deployed to Vercel be covered? I know most of these apps will be deployed there, but wouldn't be safer to check the `NODE_ENV`'s value?

Or if this is related to the `trustHost` configuration in [`packages/auth/src/config.ts#L35`](https://github.com/t3-oss/create-t3-turbo/pull/1054/files#diff-8643fbe71ccca0ed5101e1447adcd63902a0ea9daa4eaab4655c37ecb3875003R35), could it be reused here?

Comment (Wundero) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:19]: It looks like I can just use `NODE_ENV === "production"` and it will handle vercel preview envs just fine, so I will move to  that check instead.

apps/nextjs/src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -1,18 +1,42 @@
...
+ * @returns The modified request.
+ */
+function rewriteRequestUrl(req: NextRequest) {
+  if (env.NODE_ENV === 'production') {
+    return req;
+  }
+  const host = req.headers.get("host");
+  const newURL = new URL(req.url);
+  newURL.host = host ?? req.nextUrl.host;
+  return new NextRequest(newURL, req);
```
Comment (juliusmarminge) [inline apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:25]: Why do we need to rewrite if we disable csrf in dev?

Comment (Wundero) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:25]: Because the generated callback URLs point to `localhost:3000` no matter what. NextJS sets the `nextUrl` to the deployment URL, which is always `localhost` on development, but we need it to be the IP of the machine, i.e. `192.168.x.y` so that it can properly route the request to the oauth provider and back.

Comment (Wundero) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:25]: Having looked through authjs's source code, the callback url is generated by `const baseUrl = env.AUTH_URL ?? request.url` (pseudocode), so because we don't set `AUTH_URL`, we need to ensure `request.url` is set to the proper value.

Comment (juliusmarminge) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:25]: So this means we have to set the ip at the oauth provider? Hmm...

Comment (Wundero) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:25]: Just modifying the request URL seems to be sufficient. In my local testing, this has worked great both with and without the auth proxy server.

Comment (juliusmarminge) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:25]: Aight I'll try and test it out asap

PR #1056, opened 2024-05-28
feat: add Vercel openapi schema for autocomplete and validation

shivanshubisht:


dBianchii:
Nice. 
It seems that `silent: boolean` is deprecated. Maybe we can remove it?

shivanshubisht:
Yeah, we can but the way to do this is from vercel dashboard. Perhaps we could mention this somewhere?

juliusmarminge (approved):

juliusmarminge:
with there being no config - maybe just remove the file?

PR #1057, opened 2024-05-28
chore: update node-version to 22 and `pnpm/action-setup` to `v4`

shivanshubisht:


juliusmarminge:
I want to use a Node LTS that's supported on Lambda/Vercel so gonna close this

PR #1061, opened 2024-06-06
Fix NativeWind Typescript

dBianchii:
https://www.nativewind.dev/v4/getting-started/typescript

### DIFF
--- a/apps/expo/nativewind-env.d.ts
+++ b/apps/expo/nativewind-env.d.ts
@@ -1,0 +1,3 @@
+/// <reference types="nativewind/types" />
+
+// NOTE: This file should not be edited and should NOT be in your git ignore


dBianchii:
Before:
:angry: 
image

After:
:hugs: 
image

juliusmarminge:
Hmm this shouldn't be necessary since i have the types included from tsconfig:

https://github.com/t3-oss/create-t3-turbo/blob/main/apps/expo/tsconfig.json#L10

dBianchii:
> Hmm this shouldn't be necessary since i have the types included from tsconfig:
> 
> https://github.com/t3-oss/create-t3-turbo/blob/main/apps/expo/tsconfig.json#L10

Still getting the error even with that

dBianchii:
I just realized we have this same exact file in src/types :thinking: Closing this because I think the error has gone away for me

PR #1062, opened 2024-06-09
chore: update eslint and typescript in package.json.hbs

aidankinzett:
Generated packages are on older versions of eslint and typescript, this PR updates the dependencies in the template file to be in line with the rest of the repo


juliusmarminge (approved):
Thanks! Do you know if there's a way to get Renovate to bump in this file too?

PR #1115, opened 2024-07-19
chore: use `turbo watch` instead of `tsc --watch`

juliusmarminge:
less watchers should be more "easy" on the dev machine running them :)

- is there some way to not have to duplicate `build` and `dev` scripts?


dBianchii:
🙏

PR #1120, opened 2024-07-20
Add ts-rest spike

shawnmclean:
Hey guys,

I'm testing out an implementation of ts-rest, however, I'm running into a typings error.

Can someone check out the code and look at the error?

```
The inferred type of 'userRouter' cannot be named without a reference to '../../../../node_modules/@ts-rest/serverless/src/lib/types'. This is likely not portable. A type annotation is necessary.ts(2742)

```

### Instructions

I got it to run by:

1. `pnpm dev`
2. `pnpm dev --filter=@acme/nextjs`

Navigate to: http://[localhost:3000/api/v1/users/1](http://localhost:3000/api/v1/users/1)

### Backstory

I'm running tRPC and ts-rest side by side. But mostly going with ts-rest because I need openapi to work and the latest ts-openapi does not work with the latest trpc.

However, I don't mind running a REST and RPC system side by side in next, one for public facing API access while the other internal. I see Documenso doing something similar.


juliusmarminge:
I'd start with patching the ts-rest pcksge and exporting all types by changing this to an `export *` statement 

https://github.com/ts-rest/ts-rest/blob/0eff92067d7e3219493dadbccf3a31273c397a1a/libs/ts-rest/serverless/src/index.ts#L4

juliusmarminge:
I'm not gonna add ts-rest in this starter, no

PR #1136, opened 2024-08-06
Use PayloadCMS instead of drizzle and Expo API routes

jancbeck:
The pitch for [PayloadCMS v3](https://payloadcms.com/docs/beta/getting-started/what-is-payload) is that instead of 

```
export const Post = pgTable("post", {
  id: uuid("id").notNull().primaryKey().defaultRandom(),
  title: varchar("name", { length: 256 }).notNull(),
  content: text("content").notNull(),
  createdAt: timestamp("created_at").defaultNow().notNull(),
  updatedAt: timestamp("updatedAt", {
    mode: "date",
    withTimezone: true,
  }).$onUpdateFn(() => sql`now()`),
});

export const CreatePostSchema = createInsertSchema(Post, {
  title: z.string().max(256),
  content: z.string().max(256),
}).omit({
  id: true,
  createdAt: true,
  updatedAt: true,
});

const allPosts = return db.query.Post.findMany({
      orderBy: desc(Post.id),
      limit: 10,
    });
```

you write 

```
export const Posts: CollectionConfig = {
  slug: "posts",
  fields: [
    {
      name: "title",
      type: "text",
    },
    {
      name: "content",
      type: "textarea",
    },
  ],
};
const posts = await payload.find({
      collection: "posts",
      limit: 10,
    });
```

plus you get

> - A full Admin Panel using React server / client components, matching the shape of your data and completely extensible with your own React components
> - Automatic database schema, including direct DB access and ownership, with migrations, transactions, proper indexing, and more
> - Instant REST, GraphQL, and straight-to-DB Node.js APIs
> - Authentication which can be used in your own apps
> - A deeply customizable access control pattern
> - File storage and image management tools like cropping / focal point selection
> - Live preview - see your frontend render content changes in realtime as you update

It uses drizzle and next.js under the hood which you can use if you want.

This PR also enables Expo Router web export and uses [Expo API routes](https://docs.expo.dev/router/reference/api-routes/) for isometric tRPC route handlers (no direction connection between Expo and Next deployment).

## Roadmap

- [ ] isometric UI between native apps, Expo web and Next.js via [gluestack](https://gluestack.io/ui/docs/home/overview/introduction)
- [ ] OAuth support (see https://github.com/contentql/pin-hcms/ for reference)
- [ ] eventually [merge Next.js with Expo](https://youtu.be/djhEgxQf3Kw)?

## Known issues

- SSR in Next.js is disabled for now until OAuth support is added.
- Expo Go is currently broken because of the Expo canary dependency


ck-euan:
Doesn't Payload only support a small handful of databases at the moment though?
So if this was merged in it'd be difficult for someone to use this repo with MySQL for example

jancbeck:
Payload is [open](https://payloadcms.com/blog/relational-database-table-structure-rfc) to supporting new SQL databases and SQL Lite just landed. But to me the point of payload is more about not having to spend much time thinking about db. If you need a specific db then it's perhaps not the right solution for you but I think this updated stack would allow for very rapid development of apps.

juliusmarminge:
Hey sorry for not reviewing this

It's an interesting setup you have here, however I don't think it suits into this repo given it stems from create-t3-app and that doesn't have Payload.

PR #1144, opened 2024-08-10
chore: repo maintenance, bump deps, cleanup configs

ahkhanjani:
Tidied up config files and updated deps, node and pnpm versions.


ahkhanjani:
I don't think there is any need for `tailwind.config.ts` and its dependencies in the UI package. Can anyone please confirm this?

ahkhanjani:
@juliusmarminge When do we default to `/src/**/*.ts` and use generated `.d.ts` files like in validators package? I don't know why sometimes we would use generated `.js` files and sometimes just the `.ts` files. Could you please explain?

I have enabled `emitDeclarationOnly` for the validators package since we're not using `.js` files but I suspect I lack the required TS knowledge to understand something here.

juliusmarminge:
The declaration is mainly to speed up typescript editor performance since it's faster to scan d.ts files than inferring from .ts files.

The transpiled .js isn't really the point, especially since Next.js still transpiles on it's own. We can unify and only have `emitDeclarationOnly` on all packages if you think that's less mental overhead

ahkhanjani:
> The transpiled .js isn't really the point, especially since Next.js still transpiles on it's own. We can unify and only have `emitDeclarationOnly` on all packages if you think that's less mental overhead

That would be awesome. Let's go with that

ahkhanjani:
I think we're good to go @juliusmarminge. Please review

tooling/typescript/internal-package.json
```diff
@@ -5,6 +5,7 @@
     /** Emit types for internal packages to speed up editor performance. */
     "declaration": true,
     "declarationMap": true,
+    "emitDeclarationOnly": true,
     "noEmit": false,
```
ahkhanjani, line 9: What happens if we remove `"declaration": true`? I know `"noEmit": false` is there to override the extended base config so it should probably stay.

apps/auth-proxy/eslint.config.js
```diff
@@ -1,9 +1,4 @@
 import baseConfig from "@acme/eslint-config/base";
 
 /** @type {import('typescript-eslint').Config} */
-export default [
-  {
-    ignores: [".nitro/**", ".output/**"],
-  },
```
juliusmarminge, line 7: i still think these block is valuable to have as it shows people how to ignore files even if they weren't ignored by some other source

packages/db/eslint.config.js
```diff
@@ -1,10 +1,4 @@
-import baseConfig, { restrictEnvAccess } from "@acme/eslint-config/base";
+import baseConfig from "@acme/eslint-config/base";
 
 /** @type {import('typescript-eslint').Config} */
-export default [
-  {
-    ignores: ["dist/**"],
-  },
-  ...baseConfig,
-  ...restrictEnvAccess,
```
juliusmarminge, line 9: this was removed by accident?

tooling/typescript/internal-package.json
juliusmarminge: cause we need declaration files

juliusmarminge (changes requested):
revert all the eslint configs and then we're good to merge

packages/db/eslint.config.js
ahkhanjani: Do you mean `restrictEnvAccess`? If yes then should it be there if we're not using t3-env? If no then the rest is the same as before now. Should I add `restrictEnvAccess` back?

ahkhanjani:
@juliusmarminge I did not update the React Native libraries since I didn't wanna break anything. Could you please look into that? Or should I leave them as they are?

tooling/typescript/internal-package.json
```diff
@@ -5,6 +5,7 @@
     /** Emit types for internal packages to speed up editor performance. */
     "declaration": true,
     "declarationMap": true,
+    "emitDeclarationOnly": true,
```
ck-euan, line 8: I think you may need to set noEmit to null here to properly override it. 

Otherwise we end up with the following error
error TS5053: Option 'emitDeclarationOnly' cannot be specified with option 'noEmit'.

apps/nextjs/src/env.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-restricted-properties */
```
juliusmarminge, line 1: this disable directive should still be required given we're referencing `process.env` in this file which should not be allowed with our custom lint config

packages/auth/env.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-restricted-properties */
```
juliusmarminge, line 1: same here as above

packages/db/eslint.config.js
juliusmarminge: i thought we were using t3-env here

tooling/eslint/base.js
```diff
@@ -10,29 +10,32 @@ import tseslint from "typescript-eslint";
...
-      "error",
-      {
-        name: "process",
-        importNames: ["env"],
-        message:
-          "Use `import { env } from '~/env'` instead to ensure validated types.",
-      },
-    ],
+export const restrictEnvAccess = tseslint.config(
+  { ignores: ["**/env.ts"] },
```
juliusmarminge, line 14: ah nvm you're setting a global ignore here

juliusmarminge:
> I did not update the React Native libraries since I didn't wanna break anything

Let's do that in a separate one. React Native 0.75 was just released so I'm guessing there's gonna be a new Expo update etc soon

PR #1158, opened 2024-08-23
fix: Continue executing `dev` when one task fails

mateossh:
Bug was introduced in #1115


necmettindev:
@juliusmarminge can we merge this pr?

PR #1171, opened 2024-09-02
Add `DB_PREFIX` environment variable

lacymorrow:
This allows users to add a prefix to their db tables to allow sharing a db between apps


juliusmarminge:
having this as a env var makes no sense to me. you don't change your db table names after init

PR #1177, opened 2024-09-04
Revert-to-upstream

jancbeck:
Undoing a lot of the changes and reverting to as close as possible to the original. Focus on just integrating payload.


Wundero:
I think you have made a PR to the wrong repo @jancbeck, this doesn't really fit the bill of ct3t.

jancbeck:
you are right @Wundero. I forked the repo and accidentially created the PR upstream instead of origin. Sorry!

PR #1184, opened 2024-09-13
feat: Update action.yml to use .nvmrc instead of hardcoded node version

polvallverdu:
Using this on my current setup, since it makes more sense. The .nvmrc file by default specifies 20.16, gh actions is using 20.17.


juliusmarminge (approved):
Nice one

Issue #1178, opened 2024-09-05
version conflict:  update shadcn cli usage


dmetelin:
### Provide environment information

  System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M1
    Memory: 457.36 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.6.2 - /usr/local/bin/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 10.8.1 - /usr/local/bin/npm
    pnpm: 9.9.0 - /opt/homebrew/bin/pnpm
    Watchman: 2024.07.15.00 - /opt/homebrew/bin/watchman

### Describe the bug

Update `pnpm ui-add` to use the new shadcn cli:
https://ui.shadcn.com/docs/cli

```
pnpm dlx shadcn-ui add && prettier src --write --list-different

│
│  ENOENT  Command failed with ENOENT: shadcn-ui add
│ spawn shadcn-ui ENOENT
```

Change to:
```
pnpm dlx shadcn add && prettier src --write --list-different
```

### Link to reproduction

https://ui.shadcn.com/docs/cli

### To reproduce

Update shadcn to the latest version.
Run `pnpm ui-add`

Bashamega:
I would like to work on this

Bashamega:
Should i create a pr?

soulful-ramble:
> Should i create a pr?

Yes, you can fork this repo and pull request. 😊

soulful-ramble:
#1186

PR #1186, opened 2024-09-16
update shadcn command to make sure it works

soulful-ramble:
fix this issue: #1178

### DIFF
--- a/packages/auth/src/config.ts
+++ b/packages/auth/src/config.ts
@@ -71,5 +71,6 @@ export const validateToken = async (
 };
 
 export const invalidateSessionToken = async (token: string) => {
-  await adapter.deleteSession?.(token);
+  const sessionToken = token.slice("Bearer ".length);
+  await adapter.deleteSession?.(sessionToken);
 };
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -24,7 +24,7 @@
     "format": "prettier --check . --ignore-path ../../.gitignore",
     "lint": "eslint",
     "typecheck": "tsc --noEmit --emitDeclarationOnly false",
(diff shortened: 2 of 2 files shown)


nicoheredia59:
```
√ Which components would you like to add? » accordion
✔ Checking registry.
⠋ Updating tailwind.config.ts
Something went wrong. Please check the error below for more details.
If the problem persists, please open an issue on GitHub.
```

Don't detecting tailwind config. I dont know if I'm the problem or should be fixed

nicoheredia59:
As a workaround, I would create tailwind.config.ts at UI package
```tsx
import type { Config } from "tailwindcss";

import baseConfig from "@acme/tailwind-config/web";

export default {
  content: [...baseConfig.content],
  presets: [baseConfig],
} satisfies Config;
```

By adding this, you're being able to add components without getting the previous error. Actually don't know another way to fix it.

Note that you need to add `@acme/tailwind-config/web` in your `package.json`

soulful-ramble:
> As a workaround, I would create tailwind.config.ts at UI package
> 
> ```tsx
> import type { Config } from "tailwindcss";
> 
> import baseConfig from "@acme/tailwind-config/web";
> 
> export default {
>   content: [...baseConfig.content],
>   presets: [baseConfig],
> } satisfies Config;
> ```
> 
> By adding this, you're being able to add components without getting the previous error. Actually don't know another way to fix it.
> 
> Note that you need to add `@acme/tailwind-config/web` in your `package.json`

It works. But I think we could modify config in components.json: 
```zsh
-    "config": "./tailwind.config.ts",
+    "config": "../../tooling/tailwind/web.ts",
```

Issue #1110, opened 2024-07-16
bug: superjson not transforming Dates for trpc mutations


McCulloughRT:
### Provide environment information

  System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M2
    Memory: 72.06 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.14.0 - ~/.nvm/versions/node/v20.14.0/bin/node
    npm: 10.7.0 - ~/.nvm/versions/node/v20.14.0/bin/npm
    pnpm: 9.5.0 - ~/.nvm/versions/node/v20.14.0/bin/pnpm
    bun: 1.1.12 - ~/.bun/bin/bun
    Watchman: 2024.05.06.00 - /opt/homebrew/bin/watchman

### Describe the bug

when performing a trpc mutation, the superjson transformer on the server/api side does not appear to transform serialized dates back into date objects, even though network logs show it is serializing it properly on the client side. It just stays as a string. This results in a zod error if the mutation has a zod input validator that specifies "z.date()" (as well as any other error you might expect down the line from having the wrong type). 

When using a trpc query with the same data (just an object with a date in it), superjson serializes and parses it without issue on both client and server.

I expanded the transformer initialization in /packages/api/src/trpc.ts to explicitly log the input values superjson recieves and returns in the transformer chain and it does appear that, given the same input from the client, in both queries and mutations trpc is providing the exact same data into the transformer argument, but for a reason I can't understand superjson seems to selectively transform only the queries and leaves the mutations unchanged.

### Link to reproduction

https://github.com/McCulloughRT/date-example

### To reproduce

See the linked repo, taking note of `apps/nextjs/src/app/dateTest` and `packages/api/src/router/dateTest`
Loading the page will issue a query that takes a date object as a param, validated in the router by zod to be a date successfully. Clicking the button with issue a mutation that takes a date object as a param, failing validation in the router to be a date.

juliusmarminge:
did you try repro this in a minimal trpc setting?

just trying to narrow down the search before digging in

McCulloughRT:
Here's a minimal version that _does not_ experience this issue using the trpc "getting started" as a guide, but with superjson and the package versions currently specified in the t3-turbo package.json: 

https://github.com/McCulloughRT/trpc-test

So it's something about the way they're coming together in t3-turbo?

McCulloughRT:
[...]

PR #1189, opened 2024-09-21
feat: expose TRPC transformer from api package

ochicf:
## Context

In #1110, a problem with `superjson` caused the necessity of swapping transformer to `devalue`. Code for `superjson` is spread in 4 different places:
- `packages/api/src/trpc.ts`
- `apps/nextjs/src/trpc/query-client.ts`
- `apps/nextjs/src/trpc/react.tsx`
- `apss/expo/src/utils/api.tsx`

Currently all of those places directly import `superjson`, which for this specific transformer is not super problematic (as its interface matches 1 to 1 to the expected transformer in tRPC), but it makes it inconvenient to swap to another or do any customization to its behavior, as it needs to be changed in all places.

## Proposal

Define the transformer in the api package and export it so it can be used in all places. Besides the benefit of applying DRY, it moves the entire responsability of the transformer where it belongs (close to the API) and the apps are just consumers.

### Swapping transformers

Now swapping transformers will be super easy and convenient and fully transparent to developers in their local machines, though underneath it's a big breaking change as it drastically changes how the data is being sent through the API. This is why I found it necessary to add a disclaimer in that file just in case.

For example, if there are already distributed apps with the `superjson` transformer and there's a swap to `devalue`, all of those app's API calls will break when the API using `devalue` gets released.

Of course, this would happen before as well and it's out of the scope of this repo, but IMO as now it's easier and "seems prepared to do it" adding the disclaimer made sense.

#### Including the module

To include the `@acme/api/module` I had to add it in the `tsconfig.json`'s `compilerOptions.paths` section, pointing directly to the file.

I'm not really sure if this is correct, but it's how I managed to make it work.

#### Restricting usage from `@acme/api` package

Since the `@acme/api` package is now a direct dependency in the expo app, I've added some eslint rules to prevent importing values directly from it, or from modules other than `/transformer`.

Example of imports 

```ts
import type { AppRouter } from "@acme/api"; // <- correct: imports type
import type { hello as _1 } from "@acme/api/hello"; // <- correct: imports type
import { createCaller as _2 } from "@acme/api"; // <- incorrec: imports value
import { transformer } from "@acme/api/transformer"; // <- correct: imports value from allowed module
import { world as _3 } from "@acme/api/world"; // <- incorrect: imports value from restricted module
```

```sh
[...]/create-t3-turbo/apps/expo/src/utils/api.tsx
   8:1  error  '@acme/api' import is restricted from being used. Only type imports from '@acme/api' are allowed                            @typescript-eslint/no-restricted-imports
  10:1  error  '@acme/api/world' import is restricted from being used by a pattern. Only certain modules from '@acme/api' can be imported  @typescript-eslint/no-restricted-imports

✖ 2 problems (2 errors, 0 warnings)
```

### DIFF
--- a/apps/expo/package.json
+++ b/apps/expo/package.json
@@ -17,7 +17,7 @@
   "dependencies": {
     "@bacons/text-decoder": "^0.0.0",
     "@expo/metro-config": "^0.18.11",
-    "@shopify/flash-list": "1.6.4",
+    "@shopify/flash-list": "1.7.1",
     "@tanstack/react-query": "catalog:",
     "@trpc/client": "catalog:",
     "@trpc/react-query": "catalog:",
--- a/apps/expo/package.json
+++ b/apps/expo/package.json
@@ -40,8 +40,7 @@
     "react-native-gesture-handler": "~2.16.2",
     "react-native-reanimated": "~3.10.1",
     "react-native-safe-area-context": "~4.10.8",
-    "react-native-screens": "~3.31.1",
-    "superjson": "2.2.1"
+    "react-native-screens": "~3.31.1"
   },
   "devDependencies": {
     "@acme/eslint-config": "workspace:*",
--- a/apps/expo/package.json
+++ b/apps/expo/package.json
@@ -15,6 +15,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
+    "@acme/api": "workspace:*",
     "@bacons/text-decoder": "^0.0.0",
     "@expo/metro-config": "^0.18.11",
     "@shopify/flash-list": "1.7.1",
@@ -43,7 +44,6 @@
     "superjson": "2.2.1"
   },
   "devDependencies": {
-    "@acme/api": "workspace:*",
     "@acme/eslint-config": "workspace:*",
     "@acme/prettier-config": "workspace:*",
     "@acme/tailwind-config": "workspace:*",
--- a/packages/api/src/transformer.ts
+++ b/packages/api/src/transformer.ts
@@ -1,0 +1,22 @@
+import type { TRPCCombinedDataTransformer } from "@trpc/server";
+import SuperJSON from "superjson";
+
+/**
+ * tRPC transformer, used to serialize/deserialize data between client and server.
+ * This export is used internally in `@acme/api` and in both `apps/nextjs` and `apps/expo`.
+ *
+ * ! IMPORTANT: changing this is a BREAKING CHANGE !
+ * ? Even though this helps swapping transformers, bear in mind that distributed Expo apps
+ * ? will have the previous transformer bundled in their device, which will cause a mismatch
+ * ? in how the data is sent and received (basically, all API requests will fail).
+ */
+export const transformer: TRPCCombinedDataTransformer = {
+  input: {
+    serialize: SuperJSON.serialize,
+    deserialize: SuperJSON.deserialize,
+  },
+  output: {
+    serialize: SuperJSON.serialize,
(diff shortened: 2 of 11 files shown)


apps/expo/package.json
```diff
@@ -15,9 +15,10 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
+    "@acme/api": "workspace:*",
```
juliusmarminge, line 18: Have Expo shipped tree shaking yet? If not I think including this as a prod dep will leak backend code

packages/api/src/transformer.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * tRPC transformer, used to serialize/deserialize data between client and server.
+ * This export is used internally in `@acme/api` and in both `apps/nextjs` and `apps/expo`.
+ *
+ * ! IMPORTANT: changing this is a BREAKING CHANGE !
+ * ? Even though this helps swapping transformers, bear in mind that distributed Expo apps
+ * ? will have the previous transformer bundled in their device, which will cause a mismatch
+ * ? in how the data is sent and received (basically, all API requests will fail).
+ */
```
juliusmarminge, line 12: good callout. add a `@see` comment to trpc docs here too as a reference. not sure many people have used the "full input/output syntax" before

apps/expo/package.json
juliusmarminge: https://x.com/anis_RNCore/status/1841790356930887859

ochicf: Good point, didn't though of that. I'm not sure how to check if it's happening (I'm quite new with Expo and RN). Also, not sure if updating SDK is easy or not and should be done in this PR or in a different one.

Another option would be to move the transformer logic into a separate package so we are 100% sure it doesn't bring any BE code. For example `@acme/transformer` or `@acme/api-transformer`

dBianchii:
Ok, so swapping the transform is for cases where the current superJSON might not be enough? When will that be the case? 
Isn't the date issue for mutations a problem with superJSON directly?

apps/expo/package.json
dBianchii: I am personally scared of adding an api package as a dependency, regardles of wether or not Expo does tree-shaking correctly or not. Might not be an issue, but it's something to be mindful of.

I personally have a `@acme/shared` package that exports common utils/stuff  that is safe to be shared everywhere. Adding another package just for the tranformer export might be a bit verbose for this repo though

ochicf: Cool, yeah it makes sense. I'll prepare that and revert the api package to just dev dependency in Expo.

ochicf:
@dBianchii 
> Ok, so swapping the transform is for cases where the current superJSON might not be enough? When will that be the case?

Yes. Though bear in mind that this refactor ends up with having the transformer in a single place, which would allow you to easily/consistently enhance superjson (for example to perform serialisation for custom entities, see [recipes section in superjson documentation](https://github.com/flightcontrolhq/superjson?tab=readme-ov-file#recipes)).

> Isn't the date issue for mutations a problem with superJSON directly?

In this case yes. For this specific issue, having it a single place would allow to to do a local fix on top of superjson's deserialise function, at least as a temporary fix while the fix is done at their end (this was the first approach I did in my project before swapping to `devalue`).

--------

I think having transformer in a single place is a good thing also to let developers choose the transformer they prefer, not due to any bug (for example, they may like more [devalue due to performance/footprint](https://github.com/Rich-Harris/superjson-and-devalue?tab=readme-ov-file))

PR #1227, opened 2024-10-21
Improve dev script

dBianchii:
This PR enhances the dev script by adding a dependency on the `dev` task.

Before this change, if I edited something for example in the database schema, it would only trigger changes in `@acme/db`, and it would not trigger tsc for `@acme/api`. You would need to manually trigger it with another process (eg. `pnpm build -F @acme/api...`). Now, any changes to a package will correctly cascade down all tscs for packages that depend on it

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -10,6 +10,7 @@
       "outputs": [".cache/tsbuildinfo.json", "dist/**"]
     },
     "dev": {
+      "dependsOn": ["^dev"],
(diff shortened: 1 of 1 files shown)


VaniaPopovic:
This looks like a very good change! I had to manually stop the server and rerun the dev server for eg. db changes to propagate to the nextjs app

juliusmarminge (approved):

PR #1233, opened 2024-10-25
Next 15 + React 19

ahkhanjani:


jpainam:
Thanks for the migration. 

i'm was able to `pn build`, `pn lint` and `pn typecheck` locally, but when i open `localhost:3000`, i get a lot of  `module not found errors`

```tsx
Module not found: Can't resolve @acme/ui/button
```

for every single import from `packages`.

Anything i'm missing out?

`alias pn=pnpm`

ahkhanjani:
> i'm was able to `pn build`, `pn lint` and `pn typecheck` locally, but when i open `localhost:3000`, i get a lot of `module not found errors`

Should've been fixed. I'll look into it

ahkhanjani:
@jpainam Works fine for me. Make sure you've updated the `exports` field in `packages/ui/package.json`. Turpopack doesn't work with wildcard `**/*` exports.

apps/nextjs/next.config.ts
```diff
@@ -1,13 +1,8 @@
-import { fileURLToPath } from "url";
-import { createJiti } from "jiti";
+import type { NextConfig } from "next";
 
-// Import env files to validate at build time. Use jiti so we can load .ts files in here.
-await createJiti(fileURLToPath(import.meta.url)).import("./src/env");
+import "./src/env";
```
juliusmarminge, line 3: Until next supports ESM in ts configs, I prefer to keep this .js

package.json
```diff
@@ -25,9 +25,9 @@
   },
   "devDependencies": {
     "@acme/prettier-config": "workspace:*",
-    "@turbo/gen": "^2.1.3",
+    "@turbo/gen": "~2.1.3",
     "prettier": "catalog:",
-    "turbo": "^2.1.3",
+    "turbo": "~2.1.3",
```
juliusmarminge, line 30: ??

apps/expo/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "baseUrl": ".",
     "paths": {
-      "~/*": ["./src/*"]
+      "~/*": ["./src/*"],
+      "react": ["./node_modules/@types/react"]
```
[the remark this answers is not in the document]
ahkhanjani: Stopped hoisting `@types/react` because expo is still on React 18. The only hack I found to make expo use React 18 types

apps/auth-proxy/package.json
```diff
@@ -10,14 +10,14 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@auth/core": "0.34.2"
+    "@auth/core": "0.37.2"
```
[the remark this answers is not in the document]
ahkhanjani: I'll take a look

apps/nextjs/next.config.ts
ahkhanjani: It doesn't work? I tested and worked just fine. Is there any other reason?

package.json
ahkhanjani: Turbo 2.2 crashes my projects in my work repo. I think it also did in this one. Not sure. I'll try again

apps/nextjs/next.config.ts
juliusmarminge: Try doing something ESM like accessing import.meta or similar.

package.json
ahkhanjani: Yup. Also panics here

jpainam:
> @jpainam Works fine for me. Make sure you've updated the `exports` field in `packages/ui/package.json`. Turpopack doesn't work with wildcard `**/*` exports.

That was the issue. Had to export each file individually

juliusmarminge:
> @jpainam Works fine for me. Make sure you've updated the `exports` field in `packages/ui/package.json`. Turpopack doesn't work with wildcard `**/*` exports.

Is that a bug or intended behavior ?

ahkhanjani:
> Is that a bug or intended behavior ?

I couldn't find anything mentioning this directly. There is [this indirect mention](https://github.com/vercel/turborepo/issues/6844#issuecomment-1866372507). I think they simply don't support it.

juliusmarminge:
That's for turborepo though, I'm guessing the limitation you're pointing at is about turbopack?

Either way I'm fine with explicit exports just thought if we should file an issue to turbopack since afaik everything that works with webpack should work in turbopack too

ahkhanjani:
> I'm guessing the limitation you're pointing at is about turbopack?

Yes

> Either way I'm fine with explicit exports just thought if we should file an issue to turbopack since afaik everything that works with webpack should work in turbopack too

That's fair. I'm going to open an issue on vercel/next.js.

This aside though, it sucks to manually add the export for each ui element every time you install it but having autocomplete is worth it imo. Should we keep it explicit either way?

juliusmarminge:
> This aside though, it sucks to manually add the export for each ui element every time you install it but having autocomplete is worth it imo. Should we keep it explicit either way?

I honestly thought we already were using explicit due to the limitation in auto-imports.

ahkhanjani:
> I honestly thought we already were using explicit due to the limitation in auto-imports.

I made the commit that changed that due to us emitting type defs and the exports getting out of control. I'm dropping `dev` and `build` scripts inside ui package now for this reason.

juliusmarminge:
> I made the commit that changed that due to us emitting type defs and the exports getting out of control. I'm dropping dev and build scripts inside ui package now for this reason.

Hmm not sure I like that reasoning 😅 There's scripting that can be done to automate it.

But either way declaration emitting is to reduce the work the language server has to do and I'm not sure these React components are very heavy to infer... It's more for Zod, tRPC, Drizzle etc that does a lot of type shenanigans that you can notice the gains of declaration files vs live inference

ahkhanjani:
> Either way I'm fine with explicit exports just thought if we should file an issue to turbopack since afaik everything that works with webpack should work in turbopack too

Turns out webpack is the problem:
https://github.com/evanw/esbuild/issues/2974#issuecomment-2303553396
https://github.com/webpack/enhanced-resolve/issues/400

I tried recreating the [issue](https://github.com/t3-oss/create-t3-turbo/pull/1233#issuecomment-2440176006) without using an array. This was what we were doing:

```json
{
  "exports": {
    "./*": [
      "./src/*.ts",
      "./src/*.tsx"
    ]
  }
}
```

I just accidentally figured out that it works just fine without the array:

```json
{
  "exports": {
    "./*": "./src/*.tsx"
  }
}
```

This is not a turbo issue. We could continue using the wildcard export for convenience and lose autocomplete (note that autocomplete works fine after importing an element manually the first time in a .ts file).

What should we do?

ahkhanjani:
> Hmm not sure I like that reasoning 😅 There's scripting that can be done to automate it if that's the main concern
> 
> But either way declaration emitting is to reduce the work the language server has to do and I'm not sure these React components are very heavy to infer... It's more for Zod, tRPC, Drizzle etc that does a lot of type shenanigans that you can notice the gains of declaration files vs live inference

I agree. This PR fixes that. That was my bad 😅

We're keeping explicit exports then

ahkhanjani:
> That was the issue. Had to export each file individually

@jpainam You can keep the wildcard export if it's more convenient for you. Just replace the array with the string as I explained and see if it works.

juliusmarminge:
The index file is a .ts so we need the array though (unless you change to index.tsx even for non-react stuff 😭)

ahkhanjani:
> The index file is a .ts so we need the array though (unless you change to index.tsx even for non-react stuff 😭)

`index.ts` is exported separately. The reason we had the array was for specific files like `use-toast.ts` which gets added when installing toast (or so it did. haven't tried recently)

ahkhanjani:
I should've written the complete version of the `exports` field. My bad.

```json
{
  "exports": {
    ".": "./src/index.ts",
    "./*": "./src/*.tsx"
  }
}
```

This should work ___IF___ you want to use the wildcard export. The recommendation is exporting each element separately. Also haven't figured out how to export wildcard `.ts` files yet. Probably add those manually

juliusmarminge:
Cool cool 👍👍

To make explicit exports less of a headache and risk of forgetting to add them, we could add a small script wrapping shadcn cli that after installation,

- adds export paths in package.json
- runs formatting

ahkhanjani:
> To make explicit exports less of a headache and risk of forgetting to add them, we could add a small script wrapping shadcn cli that after installation,
> 
> * adds export paths in package.json
> * runs formatting

Totally agree

package.json
dBianchii: Yeah, it happens with me as well. 2.2.3 is a bit better, but it still happens

apps/nextjs/next.config.js
```diff
@@ -1,13 +1,10 @@
 import { fileURLToPath } from "url";
-import { createJiti } from "jiti";
+import createJiti from "jiti";
 
-// Import env files to validate at build time. Use jiti so we can load .ts files in here.
-await createJiti(fileURLToPath(import.meta.url)).import("./src/env");
+createJiti(fileURLToPath(import.meta.url))("./src/env");
```
necmettindev, line 4: I think we don't need jiti anymore. (You can use next.config.ts directly)

juliusmarminge: https://github.com/t3-oss/create-t3-turbo/pull/1233#discussion_r1818190991

ahkhanjani:
Encountered a potential pnpm bug while trying to fix expo
https://github.com/pnpm/pnpm/issues/8715

dBianchii:
Shadcn will also update some of its components in general preparation for React 19:
https://ui.shadcn.com/docs/react-19

Not sure if they have updated it, just dropping the link here :+1:

ahkhanjani:
> Shadcn will also update some of its components in general preparation for React 19: https://ui.shadcn.com/docs/react-19
> 
> Not sure if they have updated it, just dropping the link here 👍

I couldn't find anything we can change in the page you mentioned. Seems like something for third-party cli packages. Am I missing something?

ahkhanjani:
> Shadcn will also update some of its components in general preparation for React 19: https://ui.shadcn.com/docs/react-19
> 
> Not sure if they have updated it, just dropping the link here 👍

> Can remove all the forwardRef stuff but let's do that in a follow up PR

#1243

bryanjtc:
Maybe you can reduce the pr size by reducing dependency version changes and moving bug fixes to other pull requests. The turbo issue seems to be fixed in 2.3.0

ahkhanjani:
> Maybe you can reduce the pr size by reducing dependency version changes and moving bug fixes to other pull requests.

We're actually only blocked by a pnpm bug atm

ahkhanjani:
> Maybe you can reduce the pr size by reducing dependency version changes and moving bug fixes to other pull requests.

Sorry it appears that I misread your comment. Most of the dependency updates are necessary imo because they add React 19 or Next 15 support. For example next-auth awaits for cookies etc.

juliusmarminge:
I'll try and find some time to check if latest next auth works with our expo setup, if so we can bump that on next 14 - but I ran into some weird issues when updating last time :/

ahkhanjani:
@juliusmarminge I think injecting auth as with react 18 will solve the type issue. Using react 18 types globally would be a little...

juliusmarminge:
> @juliusmarminge I think injecting auth as with react 18 will solve the type issue. Using react 18 types globally would be a little...

i don't think we'll be able to merge this PR until we've added expo 52 cause using different versions of react in a monorepo with hoisting is hard to get working from my experience...

ahkhanjani:
> i don't think we'll be able to merge this PR until we've added expo 52 cause using different versions of react in a monorepo with hoisting is hard to get working from my experience...

What if we just go with react 18 for now?

juliusmarminge:
> > i don't think we'll be able to merge this PR until we've added expo 52 cause using different versions of react in a monorepo with hoisting is hard to get working from my experience...
> 
> What if we just go with react 18 for now?

even if you have r18 in package.sjon next app dir will use r19 - so you're just "masking" the issue

czystyl:
A potential solution to have React 19 inside the Expo too: 
https://x.com/Baconbrix/status/1864785257297305833?t=nyfQ4F7CztTCXhUdYEilBQ&s=19

.nvmrc
```diff
@@ -1 +1 @@
-20.18
+22.11
```
czystyl, line 1: Node v22 causes prettier plugin warnings: 
image

Prettier v3.4 solves this issue. Could you bump the prettier version?

bryanjtc:
https://reactnative.dev/blog/2025/02/19/react-native-0.78
Not long until expo uses react 19.

juliusmarminge:
Lets do it all in #1312 - your commits have been pulled over there

PR #1237, opened 2024-10-30
Ideas about pnpm catalog protocol

ahkhanjani:
Treat this as a proposal. What are your opinions on this?

- Use YAML's anchors/references for namespace packages that must use the same version period (could include turbo as well).
- Add the frequently used `@types/node` using only the major version. Could also be sepcified as `20.x`.


pnpm-workspace.yaml
```diff
@@ -4,11 +4,12 @@ packages:
...
 catalog:
-  "@tanstack/react-query": ^5.59.15
-  "@trpc/client": ^11.0.0-rc.477
-  "@trpc/react-query": ^11.0.0-rc.477
-  "@trpc/server": ^11.0.0-rc.477
-  eslint: ^9.12.0
+  "@tanstack/react-query": ^5.59.16
+  "@trpc/client": &trpcVersion ^11.0.0-rc.601
+  "@trpc/react-query": *trpcVersion
+  "@trpc/server": *trpcVersion
```
juliusmarminge, line 10: ehh idk... i get the idea but it just looks weird... i prefer just separating groups with a newline or something to make it clear they're related?

ahkhanjani: > ehh idk... i get the idea but it just looks weird... i prefer just separating groups with a newline or something to make it clear they're related?

Something like this?

```yaml
```

PR #1243, opened 2024-11-09
shadcn React 19

ahkhanjani:
Update shadcn components to use React 19 (removed forwardRefs) and clean up.


ahkhanjani:
This should merge after #1233

packages/ui/src/button.tsx
```diff
@@ -36,23 +36,25 @@ const buttonVariants = cva(
...
-    return (
-      <Comp
-        className={cn(buttonVariants({ variant, size, className }))}
-        ref={ref}
-        {...props}
-      />
-    );
-  },
-);
-Button.displayName = "Button";
```
dBianchii, line 56: Hi, I'm curious as to why we don't need displayName anymore. (tbh I don't know the purpose of it in the first place) do you have any resources to share about this?

ahkhanjani: Yes. Same goes for `memo()`.

ahkhanjani: > (tbh I don't know the purpose of it in the first place)

https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/display-name.md
> DisplayName allows you to name your component. This name is used by React in debugging messages.

juliusmarminge (approved):

PR #1277, opened 2025-01-04
remove confusing links from github issue templates

maxwiseman:
I've been using this template for a while for a little project I'm working on. I had an idea for a feature that I didn't want to forget, so I figured I'd quickly make an issue on Github. I went to the issues tab of **my own repository**, clicked new issue, clicked the feature request template (or so I thought), and filled out the form. 

Turns out, I created a discussion post (#1276) on **this repository** without realizing it. It's because of a .github config file that uses hard coded links back to this repository. There's no mention in the README about it, so I didn't know to change it when I made my copy of the template. I don't seem to be the only one confused by this. In fact, the discussion post created just before mine was caused by the same weird link (#1275). 

Removing this file should resolve that. Obviously it will also remove these two links from this repo, but (at least from my perspective) that isn't a huge loss. **This does not remove the discussion templates themselves. Only the links to them from the new issue screen.**
2025-01-03 at 22 11 33@2x

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,7 +1,0 @@
-contact_links:
-  - name: Ask a question
-    url: https://github.com/t3-oss/create-t3-turbo/discussions
-    about: Ask questions and discuss with other community members
-  - name: Feature request
-    url: https://github.com/t3-oss/create-t3-turbo/discussions/new?category=ideas
-    about: Feature requests should be opened as discussions


juliusmarminge:
I want them there otherwise people start posting them as issues when it's not clear where they should go

maxwiseman:
@juliusmarminge That's understandable, but I feel like there is probably still another way to solve this.

For example, you could add a markdown callout like in the image, and/or set `blank_issues_enabled` to `false` ([ex](https://github.com/maxwiseman/create-t3-turbo/issues/new/choose)). Looking through the issues it seems like they're all either bug reports, questions, or feature requests, so disabling `blank_issues_enabled` might do the job.
2025-01-04 at 20 31 35@2x

Could also just add a line to the readme reminding people to change those links.
If you're open to any of these solutions I'm happy to either update this PR or make a new one.

PR #1281, opened 2025-01-06
fix: disable edge runtime by default to enable windows development

### LINKED ISSUE #1280: bug: Internal Server Error when running nextjs app locally on Windows

```
System:
    OS: Windows 10 10.0.19045
    CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
    Memory: 35.56 GB / 63.91 GB
  Binaries:
    Node: 22.12.0 - C:\nvm4w\nodejs\node.EXE
    npm: 10.9.0 - C:\nvm4w\nodejs\npm.CMD
    pnpm: 9.15.2 - C:\nvm4w\nodejs\pnpm.CMD
    bun: 1.1.40 - ~\.bun\bin\bun.EXE
```

### Describe the bug

From a fresh clone of this repo (on a windows machine), I ran `pnpm i`, set my DB URL, and then ran `pnpm dev`. 

When I load `localhost:3000` in the browser, I see "Internal Server Error" rendered, and the nextjs app logs look like this:

```
│>>> tRPC Request from rsc by undefined
│⨯ Internal error: TypeError: Cannot read properties of undefined (reading 'default')
at resolveClientReference (../../node_modules/next/dist/compiled/react-server-dom-webpack/cjs/react-server-dom-webpack-client.edge.development.js:87:55)
│    at resolveModule (../../node_modules/next/dist/compiled/react-server-dom-webpack/cjs/react-server-dom-webpack-client.edge.development.js:1667:35)       
│    at processFullRow (../../node_modules/next/dist/compiled/react-server-dom-webpack/cjs/react-server-dom-webpack-client.edge.development.js:1740:25)      
│    at processBinaryChunk (../../node_modules/next/dist/compiled/react-server-dom-webpack/cjs/react-server-dom-webpack-client.edge.development.js:1858:21)  
│    at progress (../../node_modules/next/dist/compiled/react-server-dom-webpack/cjs/react-server-dom-webpack-client.edge.development.js:1926:17)
│    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
│digest: "4207119005"
│[TypeError: Cannot read properties of undefined (reading 'default')]
│ ⨯ [TypeError: Cannot read properties of undefined (reading 'default')]
│ ⨯ [TypeError: Cannot read properties of null (reading 'default')]
│ GET / 500 in 1804ms
```

Doing the exact same thing on my Mac does not show an error and the page renders correctly.

### Link to reproduction

fresh clone of this repo

### To reproduce

From a fresh clone of this repo, I ran `pnpm i`, set my DB URL, ran `pnpm db:push`, and then ran `pnpm dev`.  

This is on windows, using Powershell (in VS Code) with nvm-windows installed.

### Additional information

This appears to be a windows-only issue, but not sure why other users on Windows haven't reported it. For me, the project is entirely unusable for development on Windows.

nickreynolds:
* Because of the bug described [here](https://github.com/vercel/next.js/issues/53562), I updated the line that exports runtime to check if the process is being run on Windows and select the nodejs runtime in that case.
* I also updated the README to explain this, since it could otherwise be confusing, and my changes should be removed if that bug is fixed.
* fixes #1280

### DIFF
--- a/apps/nextjs/src/app/api/auth/[...nextauth]/route.ts
+++ b/apps/nextjs/src/app/api/auth/[...nextauth]/route.ts
@@ -3,8 +3,6 @@ import { NextRequest, NextResponse } from "next/server";
 
 import { handlers, isSecureContext } from "@acme/auth";
 
-// Uncomment this line to enable edge runtime if your environment supports it
-// export const runtime = "edge";
 
 const EXPO_COOKIE_NAME = "__acme-expo-redirect-state";
 const AUTH_COOKIE_PATTERN = /authjs\.session-token=([^;]+)/;
--- a/apps/nextjs/src/app/api/auth/[...nextauth]/route.ts
+++ b/apps/nextjs/src/app/api/auth/[...nextauth]/route.ts
@@ -3,7 +3,6 @@ import { NextRequest, NextResponse } from "next/server";
 
 import { handlers, isSecureContext } from "@acme/auth";
 
-
 const EXPO_COOKIE_NAME = "__acme-expo-redirect-state";
 const AUTH_COOKIE_PATTERN = /authjs\.session-token=([^;]+)/;
 
--- a/apps/nextjs/src/app/api/auth/[...nextauth]/route.ts
+++ b/apps/nextjs/src/app/api/auth/[...nextauth]/route.ts
@@ -3,7 +3,8 @@ import { NextRequest, NextResponse } from "next/server";
 
 import { handlers, isSecureContext } from "@acme/auth";
 
-export const runtime = "edge";
+// Uncomment this line to enable edge runtime if your environment supports it
+// export const runtime = "edge";
 
 const EXPO_COOKIE_NAME = "__acme-expo-redirect-state";
 const AUTH_COOKIE_PATTERN = /authjs\.session-token=([^;]+)/;
--- a/apps/nextjs/src/app/api/trpc/[trpc]/route.ts
+++ b/apps/nextjs/src/app/api/trpc/[trpc]/route.ts
@@ -3,7 +3,6 @@ import { fetchRequestHandler } from "@trpc/server/adapters/fetch";
 import { appRouter, createTRPCContext } from "@acme/api";
 import { auth } from "@acme/auth";
 
-
 /**
  * Configure basic CORS headers
  * You should extend this to match your needs
--- a/apps/nextjs/src/app/api/trpc/[trpc]/route.ts
+++ b/apps/nextjs/src/app/api/trpc/[trpc]/route.ts
@@ -3,7 +3,8 @@ import { fetchRequestHandler } from "@trpc/server/adapters/fetch";
 import { appRouter, createTRPCContext } from "@acme/api";
 import { auth } from "@acme/auth";
 
-export const runtime = "edge";
+// Uncomment this line to enable edge runtime if your environment supports it
+// export const runtime = "edge";
 
 /**
  * Configure basic CORS headers
--- a/apps/nextjs/src/app/api/trpc/[trpc]/route.ts
+++ b/apps/nextjs/src/app/api/trpc/[trpc]/route.ts
@@ -3,8 +3,6 @@ import { fetchRequestHandler } from "@trpc/server/adapters/fetch";
 import { appRouter, createTRPCContext } from "@acme/api";
 import { auth } from "@acme/auth";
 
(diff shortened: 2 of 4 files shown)


apps/nextjs/src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -3,7 +3,7 @@ import { NextRequest, NextResponse } from "next/server";
 
 import { handlers, isSecureContext } from "@acme/auth";
 
-export const runtime = "edge";
+export const runtime = process.platform === "win32" ? "nodejs" : "edge";
```
Comment (juliusmarminge) [inline apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:6]: Pretty sure This doesn't work as this string need to be statically analyzavlr?

We can just remove it though if it's causing issues?

Comment (nickreynolds) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:6]: it appears to work for me locally (i.e. I'm able to develop on Windows with this change) and my vercel deployments seem fine as well (although I'm not sure if they're using the correct runtime now).

As far as I can tell there are 3 options:
1. update runtime export as done in this PR
2. remove runtime export from all pages
3. leave all code as-is and just add note for Windows developers

I defer to you on the best approach moving forward though, and happy to update this PR at your discretion.

Comment (juliusmarminge) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:6]: > it appears to work for me locally (i.e. I'm able to develop on Windows with this change) and my vercel deployments seem fine as well (although I'm not sure if they're using the correct runtime now).

if this is the case, it means you're using different runtime in prod vs dev which sounds scary. I prefer no. 2

Comment (nickreynolds) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:6]: Good point. If you think that having windows-compatibility-by-default is more important than edge-runtime-by-default, I can do no. 2. Otherwise, I'd do no. 3. Do you want me to proceed with option 2?

Comment (nickreynolds) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:6]: yes, I was incorrect about my original solution working, my IDE just wasn't reporting the error, and the line was being ignored entirely (which meant it was functionally equivalent to deleting the line, at least for local dev, but obviously not the right solution)

Comment (juliusmarminge) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:6]: > Do you want me to proceed with option 2?

Yes

Comment (nickreynolds) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts]: @juliusmarminge I made the changes to this PR that we discussed. Do you think we need to update the `drizzle.config.ts` as well? I'm not particularly familiar with this, but should we switch to a pooled DB URL now that we aren't using edge runtime?

Comment (juliusmarminge) [reply apps/nextjs/src/app/api/auth/[...nextauth]/route.ts]: you don't want a pooled url when running migrations/schema changes

apps/nextjs/src/app/api/trpc/[trpc]/route.ts
```diff
@@ -3,7 +3,8 @@ import { fetchRequestHandler } from "@trpc/server/adapters/fetch";
 import { appRouter, createTRPCContext } from "@acme/api";
 import { auth } from "@acme/auth";
 
-export const runtime = "edge";
+// Uncomment this line to enable edge runtime if your environment supports it
+// export const runtime = "edge";
```
Comment (juliusmarminge) [inline apps/nextjs/src/app/api/trpc/[trpc]/route.ts:7]:
```suggestion
```

i'd rather just remove it. people who need and want it will know how to enable it

apps/nextjs/src/app/page.tsx
```diff
@@ -8,7 +8,8 @@ import {
   PostList,
 } from "./_components/posts";
 
-export const runtime = "edge";
+// Uncomment this line to enable edge runtime if your environment supports it
+// export const runtime = "edge";
```
juliusmarminge, line 12:
```suggestion
```

apps/nextjs/src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -3,7 +3,8 @@ import { NextRequest, NextResponse } from "next/server";
 
 import { handlers, isSecureContext } from "@acme/auth";
 
-export const runtime = "edge";
+// Uncomment this line to enable edge runtime if your environment supports it
+// export const runtime = "edge";
```
Comment (juliusmarminge) [inline apps/nextjs/src/app/api/auth/[...nextauth]/route.ts:7]:
```suggestion
```

juliusmarminge (approved):

PR #1284, opened 2025-01-13
minor(api): add return type into createTRPCContext for more type safety

vinhloc30796:
Just more type checking, slightly better DevEx

### DIFF
--- a/packages/api/src/trpc.ts
+++ b/packages/api/src/trpc.ts
@@ -14,6 +14,17 @@ import type { Session } from "@acme/auth";
 import { auth, validateToken } from "@acme/auth";
 import { db } from "@acme/db/client";
 
+/** Context object passed to all tRPC procedures containing session, database and auth token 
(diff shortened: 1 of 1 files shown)


danielcspaiva:
I don't think you need to explicitly type the return, since it is already inferred by the return type of the function.

vinhloc30796:
@danielcspaiva strange, intellisense didn't work for me out of the box -- it only works after adding the type explicitly

danielcspaiva:
@vinhloc30796 it might be something in your environment. This is a fresh project using t3-turbo. No explicit typing, but the type is correctly inferred from the return of the function.

Screenshot 2025-01-14 at 09 56 50

juliusmarminge:
this is somethnig in your env - do you have your typescript version selected from the workspace? going to close this as inference is preferred here. open a discussion asking for help if you can't figure it out

PR #1321, opened 2025-03-22
Feat: better-auth

VaniaPopovic:


apps/auth-proxy/package.json
juliusmarminge, line 1: how are preview deployments handled now?

apps/expo/metro.config.js
```diff
@@ -12,10 +12,6 @@ const config = withTurborepoManagedCache(
   }),
 );
 
-// XXX: Resolve our exports in workspace packages
-// https://github.com/expo/expo/issues/26926
-config.resolver.unstable_enablePackageExports = true;
```
juliusmarminge, line 17: how come this is not needed anymore?

juliusmarminge: oh lol it's added below instead xd

apps/nextjs/src/middleware.ts
```diff
@@ -1,9 +1,29 @@
...
+import type { NextRequest } from "next/server";
+import { NextResponse } from "next/server";
 
-// Or like this if you need to do something here.
-// export default auth((req) => {
-//   console.log(req.auth) //  { session: { user: { ... } } }
-// })
+//import { client } from "@acme/auth/middleware";
+
+export default function authMiddleware(_request: NextRequest) {
```
juliusmarminge, line 6: don't you need something to handle session refreshing?

apps/expo/package.json
```diff
@@ -15,12 +15,14 @@
...
   },
   "dependencies": {
+    "@better-auth/expo": "1.2.5-beta.5",
     "@expo/metro-config": "^0.19.9",
     "@shopify/flash-list": "1.7.3",
     "@tanstack/react-query": "catalog:",
     "@trpc/client": "catalog:",
     "@trpc/server": "catalog:",
     "@trpc/tanstack-react-query": "catalog:",
+    "better-auth": "1.2.5-beta.5",
```
juliusmarminge, line 25: what's in this beta that's not in 1.2.4?

packages/auth/src/auth.ts
```diff
@@ -0,0 +1,28 @@
...
+  database: drizzleAdapter(db, {
+    provider: "pg",
+  }),
+  secret: env.AUTH_SECRET,
+  plugins: [oAuthProxy(), expo()],
+  socialProviders: {
+    discord: {
+      clientId: env.AUTH_DISCORD_ID,
+      clientSecret: env.AUTH_DISCORD_SECRET,
+      redirectURI: "http://localhost:3000/api/auth/callback/discord",
```
juliusmarminge, line 21: testing?

packages/ui/package.json
```diff
@@ -22,14 +22,14 @@
     "ui-add": "pnpm dlx shadcn@latest add && prettier src --write --list-different"
   },
   "dependencies": {
-    "@hookform/resolvers": "^3.10.0",
+    "@hookform/resolvers": "^3.1.0",
```
juliusmarminge, line 25: ?

apps/expo/package.json
VaniaPopovic: There's a small bugfix for typings in the expo client, specifically https://github.com/better-auth/better-auth/issues/1855

packages/auth/src/auth.ts
VaniaPopovic: This should ideally be changed to the proxy url when deployed

apps/nextjs/src/middleware.ts
VaniaPopovic: image
[Should update by itself](https://www.better-auth.com/docs/concepts/session-management)

packages/auth/src/auth.ts
juliusmarminge: what's the proxy URL?

juliusmarminge: can we use [`env.VERCEL_PROJECT_PRODUCTION_URL`](https://vercel.com/docs/environment-variables/system-environment-variables#VERCEL_PROJECT_PRODUCTION_URL)

apps/nextjs/src/middleware.ts
juliusmarminge: not sure how that works in next.js where you can only set headers in server actions and middleware though?

when a user is browsing pages they won't be refreshing their session, right

packages/auth/src/auth.ts
VaniaPopovic: Yes, will change it now

apps/nextjs/src/middleware.ts
Bekacru: The session is refreshed whenever /get-session is called, whether it's from a server action, middleware, or from the client. Only server component calls won't be able to do this because they can't set cookies. But, the session in the database will still be refreshed.

juliusmarminge: yea so I'm right in assuming that if we don't have a middleware in this repo, the session wouldn't be refreshed since we're only making authApi calls from RSCs?

Bekacru: yes. I'm assuming when there's a trpc call where `getSession` is called to check for the session, it'll be refreshed. But we can also have the middleware. And, if the project can support Next.js 15.2 and above, we can switch to `auth.api.getSession`. Which will remove the API call and uses cookie caching, to prevent db calls until the cache is invalid

https://www.better-auth.com/docs/integrations/next#for-nextjs-release-1520-and-above

packages/auth/package.json
```diff
@@ -4,26 +4,25 @@
   "private": true,
   "type": "module",
   "exports": {
-    ".": {
-      "react-server": "./src/index.rsc.ts",
-      "default": "./src/index.ts"
-    },
+    ".": "./src/index.ts",
+    "./middleware": "./src/middleware.ts",
```
contrun, line 8: I can't find the `middleware.ts` file. Did I miss something?

packages/auth/src/index.ts
```diff
@@ -1,15 +1,37 @@
...
+        clientId: options.discordClientId,
+        clientSecret: options.discordClientSecret,
+        redirectURI: `${options.baseUrl}/api/auth/callback/discord`,
+      },
+    },
+    trustedOrigins: ["expo://"],
+  } satisfies BetterAuthOptions;
 
-export { handlers, auth, signIn, signOut };
+  return betterAuth(config);
```
contrun, line 33: Noob question here, [better auth docs](https://www.better-auth.com/docs/integrations/next#server-action-cookies) says we need

> When you call a function that needs to set cookies, like signInEmail or signUpEmail in a server action, cookies won’t be set. This is because server actions need to use the cookies helper from Next.js to set cookies.
> To simplify this, you can use the nextCookies plugin, which will automatically set cookies for you whenever a Set-Cookie header is present in the response.

```typescript
import { betterAuth } from "better-auth";
import { nextCookies } from "better-auth/next-js";
 
export const auth = betterAuth({
    //...your config
    plugins: [nextCookies()] // make sure this is the last plugin in the array
})
```

Why didn't we use the `nextCookies` plugin here?

packages/auth/env.ts
```diff
@@ -1,18 +1,19 @@
...
-        ? z.string().min(1)
-        : z.string().min(1).optional(),
-    NODE_ENV: z.enum(["development", "production"]).optional(),
-  },
-  client: {},
-  experimental__runtimeEnv: {},
-  skipValidation:
-    !!process.env.CI || process.env.npm_lifecycle_event === "lint",
-});
+export function authEnv() {
```
dBianchii, line 4: why is this a function and not just env = createEnv?

packages/auth/src/index.ts
```diff
@@ -1,15 +1,37 @@
...
+  discordClientId: string;
+  discordClientSecret: string;
+}) {
+  const config = {
+    database: drizzleAdapter(db, {
+      provider: "pg",
+    }),
+    baseURL: options.baseUrl,
+    secret: options.secret,
+    plugins: [oAuthProxy(), expo()],
```
asobirov, line 22: While experimenting with Better Auth and expo locally, I noticed that the `oAuthProxy` plugin seems to prevent session cookies from being properly sent to expo. This might be something worth looking into, though I'm not entirely sure whether it’s an issue with my setup or a potential bug. Removing `oAuthProxy` solved the auth issue within expo in localhost

Steps to reproduce:
- clone https://github.com/asobirov/t3-turbo-better-auth
- add `oAuthProxy()` plugin 
- Press "Login with Discord" and login
- You still stay logged out

distartin:
Any reason why t3 is moving to better-auth?
I think @authjs 5 will get a lot of new good updates

CarlosZiegler:
> Any reason why t3 is moving to better-auth? I think @authjs 5 will get a lot of new good updates

Did you tried better-auth ? DX is really great! Easy setup, and some amazing plugins/ integrations.

MendyLanda:
> Any reason why t3 is moving to better-auth?
> 
> I think @authjs 5 will get a lot of new good updates

Happy path for RN is enough a reason imo

Wundero:
> Any reason why t3 is moving to better-auth? I think @authjs 5 will get a lot of new good updates

I can think of at least 3 reasons why better-auth is a better fit for this project than authjs:
1. Better Auth has a built-in oauth proxy, circumventing the need for a separate deployment
2. Better Auth has a proper Expo integration, which is something previously hand-written for Authjs
3. Better Auth has a robust plugin suite that allows users of this template to make more diverse applications, and even make your own custom plugins. With AuthJS, integration features such as organizations, multi-session auth, password authn and mfa, etc. takes a significant amount of work to accomplish. 

It's also worth mentioning that better auth seems to have a faster release cycle, but that's always subject to change and not always that important either, so I don't really count it as a reason.

AlexKovynev:
Great! But when it will be merged?

Wundero:
> Great! But when it will be merged? And look like better need to remove NativeWind and upgrade to Tailwind 4.1.

I think changes to tailwind for expo are out of scope of this PR. I am also not confident that removing nativewind is supported, even with expo 53 and tailwind 4.1, as I couldn't find any docs indicating that nativewind is unneeded.

AlexKovynev:
I am about what can be done further. NativeWind is on Tailwind 3 and most likely is dead now.

Wundero:
> I am about what can be done further. NativeWind is on Tailwind 3 and most likely is dead now.

If you want to help remove nativewind and move to tailwind 4, id recommend opening a separate PR with those changes, since it helps review processes and also allows it to be merged separately.

packages/auth/src/index.ts
rrdesignweb: Hi thanks for putting this together -  https://github.com/asobirov/t3-turbo-better-auth
I seemed to be able to get this working with no issues using better auth - including oAuthProxy()

I did use ngrok though for forwarding localhost:3000 - otherwise I could not get it to work

Changed 
`trustedOrigins: [getBaseUrl(), "better-auth-expo://", "https://xxxx-xx-xx-xxx-xx.ngrok.io"]`

`createAuthClient({
  baseURL: "https://xxxx-xx-xx-xxx-xx.ngrok.io",`
  
 If this is what we have to do - should the docs be updated to include these instructions?

CarlosZiegler:
> I am about what can be done further. NativeWind is on Tailwind 3 and most likely is dead now.

Redesigned website and looks like will ship V5 soon. 
https://github.com/nativewind/nativewind/discussions/1422

ayush-kumar-anand:
is this working fine?? i want to use this in a project so can i use this branch??

PR #1378, opened 2025-07-06
Rename `internal-package` to `compiled-package`

dBianchii:
The thinking behind this is all packages are already internal. This is more in line with TurboRepo's own documentation and is more descriptive.

https://turborepo.com/docs/core-concepts/internal-packages#just-in-time-packages
https://turborepo.com/docs/core-concepts/internal-packages#compiled-packages


juliusmarminge (approved):
seems reasonable

PR #1379, opened 2025-07-07
Simplify ESLint configuration

dBianchii:
Idea here is to simplify eslint usage:
base -> react -> nextJs

Then, the user can just select just one of these three for their package's lint rules. Much simpler.

I also removed some redundant ignores, as they are listed in the .gitignore and `base.js`uses `includeIgnoreFile(path.join(import.meta.dirname, "../../.gitignore"))`


juliusmarminge:
i kinda like them being granular but could be convinced either way

PR #1457, opened 2025-10-26
experiment: pnpm->bun

juliusmarminge:
# EXPERIMENTAL DO NOT USE

_For now this is not intended to be merged, just trying it out_

I'll post comments throughout this PR what pain points I've experienced trying to migrate this repo to use bun


juliusmarminge:
### Configuration

- LSP plugin is outdated? I don't understand TOML so I don't even know where it's sourcing the options from, why not a `.json` file with `$schema` field pointing at the correctly versioned schema?

  image

- I don't know where to put configs? I very much like that pnpm takes all their config in `pnpm-workspace.yaml`. With bun, I have `workspaces`, `catalog`, `catalogs`, `overrides` in `package.json`, whereas `linkWorkspacePackages` in `bunfig.toml`. Would prefer all of these to be in bunfig.

juliusmarminge:
Now I'm seeing weird module resolution errors again.

- When building the Next.js app: image `@tailwindcss/postcss` is installed here: https://github.com/t3-oss/create-t3-turbo/blob/f30ea12e227f3f7337d9d474c14c6593e6a7cc32/tooling/tailwind/package.json#L17

- When building the Tanstack Start app: image `nitro` is installed here https://github.com/t3-oss/create-t3-turbo/blob/f30ea12e227f3f7337d9d474c14c6593e6a7cc32/apps/tanstack-start/package.json#L34

mmikhan:
> There appears to be 2 versions of vite@7.1.12 installed?

It seems to be a bug to Bun v1.3.0 that's already been reported https://github.com/oven-sh/bun/issues/23615. This is indeed a blocker before migrating to Bun.

Moreover, the team seems also want to have a dedupe command. https://github.com/oven-sh/bun/issues/1343

PR #1462, opened 2025-10-26
experiment: eslint->oxlint, prettier->oxfmt

juliusmarminge:
# EXPERIMENTAL DO NOT USE

_For now this is not intended to be merged, just trying it out_

### DIFF
--- a/packages/validators/package.json
+++ b/packages/validators/package.json
@@ -13,7 +13,6 @@
     "build": "tsc",
     "clean": "git clean -xdf .cache .turbo dist node_modules",
     "dev": "tsc",
-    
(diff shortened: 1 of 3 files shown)


tooling/oxlint/base.json
juliusmarminge: This is the output from running `bunx @oxlint/migrate --type-aware base.ts`. Could probably be simplified

tooling/oxlint/nextjs.json
juliusmarminge: This is the output from running `bunx @oxlint/migrate --type-aware nextjs.ts`. Could probably be simplified

PR #1478, opened 2025-11-12
fix: use baseUrl for discord oauth redirect to support local development

bcigdemoglu:
Changed Discord OAuth redirectURI from `options.productionUrl` to `options.baseUrl` to enable proper redirect handling in local development environments.

**Problem:**
- Discord OAuth callback was using `productionUrl` which prevented local development testing
- Local redirects would fail because they were pointing to production Vercel URL which were undefined in local

**Solution:**
- Use `baseUrl` instead which correctly adapts to the current environment (local or production)
- This matches the pattern already used for the main `baseURL` configuration

**Changes:**
- Updated `packages/auth/src/index.ts` line 37


bcigdemoglu:
How can I get a review on this PR @carloschida @t3dotgg ?

PR #1502, opened 2026-01-22
chore: turborepo best practices improvements

anthonyshew:
## Summary

- Update `turbo.json` schema URLs to `turborepo.dev/schema.v2.json`
- Use `turbo run` instead of shorthand in root `package.json` scripts
- Move `.env` files from root to individual packages for clearer dependency ownership

## Changes

**Schema URL**: The old `turborepo.com/schema.json` URL is outdated.

**`turbo run`**: The shorthand `turbo <task>` is intended for interactive terminal use. Scripts should use `turbo run <task>`.

**Per-package `.env` files**: A root `.env` creates implicit coupling between packages. Moving `.env.example` to each package that needs environment variables makes it clear which packages depend on which variables and improves cache correctness.

Packages now load their own `.env` via `dotenv --` instead of `dotenv -e ../../.env --`.

## AI usage
Chat transcript: https://opncd.ai/share/0Q4XFy3y


kyh:
is there a technical reason why we don't want a single `.env`?

juliusmarminge:
@cursor remove `dotenv` dependency and use Node.js built-in env loader: https://nodejs.org/api/environment_variables.html

```
CLI Options[#](https://nodejs.org/api/environment_variables.html#cli-options)
.env files can be used to populate the process.env object via one the following CLI options:

[--env-file=file](https://nodejs.org/api/cli.html#--env-filefile)

[--env-file-if-exists=file](https://nodejs.org/api/cli.html#--env-file-if-existsfile)
```

mmikhan:
I don't think this PR works at this point as expected. Even though the PR moved the env vars to packages and apps level, packages still run through the Next.js process (for Next.js app). Next.js cannot find the package level env vars unless manually referencing them in the process.

So, I had to reference package level env vars in the Next.js package.json like the following:

```
"dev": "pnpm with-env next dev --port 3000",
"with-env": "dotenv -e .env -e ../../packages/db/.env -e ../../packages/auth/.env -e ../../packages/api/.env -e ../../packages/mail/.env --",
```

anthonyshew:
👋 Just cleaning up my Pull Requests page.

